PR #19, opened 2016-09-10
Use Picasso for loading of images online.

drivfe:
Picasso will automatically load the images from the web and store it in memory/disk cache when necessary.

https://github.com/square/picasso


ram-on:
Thank you for your input.

Issue #26, opened 2016-10-07
Playback starts over after visiting channel while viewing video


atomjack:
To reproduce:

- Start watching a video, and let the video play for awhile, so that the time readout displays something other than 00:00
- Tap the i icon in the bottom left to bring up video info, then tap on the channel icon to show the Channel Browser screen.
- Tap the back arrow or back button to return to the video, and notice the video starts playing from the beginning, instead of resuming from where it had been.

(I'd be happy to work on this)

ram-on:
Thx @atomjack

ram-on:
Done.  Refer to 22a31cc55caad64c9aa976679706fd9f2a535527.

PR #28, opened 2016-10-10
When leaving the video player screen, save the current position of th…

atomjack:
…e video to resume playback when returning from the Channel Browser screen. Addresses #26.


ram-on:
I am personally not a fan of having a thread polling the video status.

Can we just set the videoCurrentPosition when onPause() is called (and then restore the video status when running onPrepared() method)?

atomjack:
Good point. I've removed the polling, and it now saves the position onPause. Also, it will restore the position if the activity is killed (I haven't tested this though).

ram-on:
Manually merged.  Refer to 22a31cc55caad64c9aa976679706fd9f2a535527.  Thx

Issue #49, opened 2017-01-31
Video Player Error.


daleylife411:
This just started today.  Any time I attempt to watch a video I get a video player error: could not load decryption code for the youtube service.  I  lowered my resolution in settings and that did not help.  Everything worked fine until now.

ZaneHannanAU:
Same issue with TeamNewPipe/NewPipe and ram-on/SkyTube.

May be related to [Chromium bug 686430: Please allow disabling Widevine/EME again][widevine] and may be adopted by YouTube for no immediately apparent reason.

[widevine]: https://bugs.chromium.org/p/chromium/issues/detail?id=686430

cmv47:
same with newpipe and skytube.decrypt error

cmv47:
please fix this

Dazrok:
same here, please fix

avently:
Here is a fix
https://github.com/TeamNewPipe/NewPipe/commit/140b480f8293353280084c634392a349c2e9b27a

theScrabi:
Right this did fix it: https://github.com/TeamNewPipe/NewPipe/commit/140b480f8293353280084c634392a349c2e9b27a#diff-7b6ba4a07fd283628d0a08013a4c27fe

atomjack:
I'm looking into fixing this on Skytube right now.

atomjack:
@ram-on, I was able to get it to work by simply prepending "https://youtube.com" to what it was using for the javascript file, and didn't need to change any of the regex stuff. I'll leave it up to you to decide if the regex changes are needed.

Dazrok:
how to apply the fix?

atomjack:
@Dazrok He released a new version with the fix, it should ask you if you want to update when you launch the app.

github-throwaway:
v2.2 is not available on F-Droid yet. You'll need to download the v2.3 apk file from [here](https://github.com/ram-on/SkyTube/releases/download/v2.2/SkyTube-Extra-2.2.apk).

ram-on:
@Dazrok Yesterday I've pushed the fixed to F-Droid.  I may take around 3 days until they compile SkyTube and produce the APK.

I suggest you download it directly from [github](https://github.com/ram-on/SkyTube/releases/download/v2.2/SkyTube-Extra-2.2.apk).

PR #50, opened 2017-01-31
Get correct url for decryption js when it doesn't include https://you…

atomjack:
…tube.com. For #49.


ram-on:
Thanks @atomjack !

Issue #61, opened 2017-02-03
[Request] Add video to bookmarks


xamica:
And possible button "see later videos" for viewing such bookmarked videos.

OffifialMITX:
That would be nice, maybe like in webTube where its a bar you swipe in from right?

atomjack:
How about a "Saved" tab (next to Featured and Most Popular, and after the Subscriptions tab which has been merged but not released yet)? I wouldn't mind implementing this. I will add an option to the new video options menu (that has copy to clipboard and share) to add to saved (or remove, if it's already saved), and then on the Saved tab, it could have a remove from saved option.

I will also see about doing drag & drop reordering.

atomjack:
How does this look?

![image](https://cloud.githubusercontent.com/assets/114672/22859135/1d52b2e6-f087-11e6-8385-6285aa39f254.png)

![image](https://cloud.githubusercontent.com/assets/114672/22859138/24dd3c2a-f087-11e6-8ba5-f392eb0cd297.png)

![image](https://cloud.githubusercontent.com/assets/114672/22859140/2995f48c-f087-11e6-9e1f-04370289b1fc.png)

Drag & Drop to reorder:
![image](https://cloud.githubusercontent.com/assets/114672/22859142/2e5c7216-f087-11e6-8448-c705efd9ffdf.png)

![image](https://cloud.githubusercontent.com/assets/114672/22859144/37886cb4-f087-11e6-96d1-41bed085af6e.png)

![image](https://cloud.githubusercontent.com/assets/114672/22859145/3aa4b95c-f087-11e6-8fc8-46f49c6002f6.png)

ram-on:
I like it @atomjack.  Well done.

ram-on:
Added the "bookmark videos" feature.  Will be released in the next release (hopefully this weekend).

Refer to commits:
* f86e6a19e4d74c05585f1c5f1807542cda3a90f4
* 3dcdc8f589ccc6ee969c8604874c90f00c677fad
* 56e34353eda03cde5909fdfb4829b2208a3b2d6d

@atomjack thx

atomjack:
[...]

PR #76, opened 2017-02-12
Added Saved Videos tab. For #61.

atomjack:


ram-on:
Please document your code by inserting comments.  I'll check out this pull request in the meantime.

atomjack:
Ok, I added some comments. I also updated a few other things, mostly to make sure scroll position on the Saved Tab is not lost when you switch back and forth (including when you add or remove a Saved video from one of the other tabs).

ram-on:
Just FYI:
- Changed the term from 'Saved' to 'Bookmarkes'.  This is to avoid ambiguity as videos are not saved (technically speaking).
 - The user is now notified whenever he bookmarkes / unbookmarkes a video (via a toast).

Issue #85, opened 2017-02-26
[Bug] An error has occurred while getting subscribed channels.


github-throwaway:
The newest version of SkyTube has no content what so ever. 
I get an error message (see screenshot) but other than that the app remains empty. So even if you search for stuff nothing shows up. 

![screenshot_2017-02-26-02-49-02](https://cloud.githubusercontent.com/assets/25488086/23336412/de898266-fbcf-11e6-9d4f-0affdcf54b13.png)
 

**N9005 | Android 5.0 | SkyTube 2.3**

SpotComms:
Its not a bug, the API key used has hit its quota, that or Google disabled it. It's something @ram-on will have to check.

MadMaverick9:
This is bad.

Interestingly when I searched for "pull the rug from under you" on ddg, the second link was: https://www.reddit.com/r/programming/comments/5mwkbz/using_google_apis_is_as_stupid_as_ever_theyll/

> Using Google API's is as stupid as ever. They'll eventually pull the rug from under you.

Whilst I much appreciate the work you are doing - I really like SkyTube, because it is better than the default youtube application - developing for a platform that you do not control seems a bad idea.

This may eventually affect applications like SMTube as well.

xamica:
Confirm problem with key.

But applications like youtube-dl will never affected by this problem, as here no keys at all.

Backtrace:
```
E/YouTubeChannel( 2018): Error has occurred while getting Featured Videos. 
E/YouTubeChannel( 2018): com.google.api.client.googleapis.json.GoogleJsonResponseException: 403 Forbidden
E/YouTubeChannel( 2018): {
E/YouTubeChannel( 2018): "code": 403,
E/YouTubeChannel( 2018): "errors": [
E/YouTubeChannel( 2018): {
E/YouTubeChannel( 2018): "domain": "usageLimits",
E/YouTubeChannel( 2018): "message": "Daily Limit Exceeded. The quota will be reset at midnight Pacific Time (PT). You may monitor your quota usage and adjust limits in the API Console: https://console.developers.google.com/apis/api/youtube/quotas?project=102659752285",
E/YouTubeChannel( 2018): "reason": "dailyLimitExceeded",
E/YouTubeChannel( 2018): "extendedHelp": "https://console.developers.google.com/apis/api/youtube/quotas?project=102659752285"
E/YouTubeChannel( 2018): }
E/YouTubeChannel( 2018): ],
[...]

PR #90, opened 2017-03-12
Added Preference for the user to enter their own YouTube Data API Key…

atomjack:
…. For #85.

### DIFF
--- a/app/src/main/java/free/rm/skytube/gui/fragments/PreferencesFragment.java
+++ b/app/src/main/java/free/rm/skytube/gui/fragments/PreferencesFragment.java
@@ -21,15 +21,24 @@ import android.app.AlertDialog;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.net.Uri;
+import android.os.AsyncTask;
 import android.os.Bundle;
+import android.preference.EditTextPreference;
 import android.preference.ListPreference;
 import android.preference.Preference;
 import android.preference.PreferenceCategory;
 import android.preference.PreferenceFragment;
+import android.widget.Toast;
+
+import com.google.api.services.youtube.YouTube;
+
+import java.io.IOException;
 
 import free.rm.skytube.BuildConfig;
 import free.rm.skytube.R;
 import free.rm.skytube.businessobjects.VideoStream.VideoResolution;
+import free.rm.skytube.businessobjects.YouTubeAPI;
+import free.rm.skytube.gui.app.SkyTubeApp;
 
 /**
  * A fragment that allows the user to change the settings of this app.  This fragment is called by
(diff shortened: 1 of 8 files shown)


ram-on:
Looks good.  I'll merge it later on. Thx

app/src/main/java/free/rm/skytube/gui/fragments/PreferencesFragment.java
```diff
@@ -104,10 +113,46 @@ public void onPause() {
...
+					new AsyncTask<Void, Void, Boolean>() {
+						@Override
+						protected Boolean doInBackground(Void... voids) {
+							try {
+								YouTube.Videos.List videosList = YouTubeAPI.create().videos().list("snippet");
+								videosList.setFields("items(id)");
+								videosList.setKey(youtubeAPIKeyPref.getText());
+								videosList.setChart("mostPopular");
+								String regionCode = SkyTubeApp.getPreferenceManager().getString(SkyTubeApp.getStr(R.string.pref_key_preferred_region), "").trim();
+								videosList.setRegionCode(regionCode.isEmpty() ? null : regionCode);
```
ram-on: Don't copy and paste code.  Just call `GetFeaturedVideos` in `AsyncTask` and perhaps modify `MAX_RESULTS` in order to retain efficiency.

app/src/main/java/free/rm/skytube/gui/fragments/PreferencesFragment.java
```diff
@@ -104,10 +113,46 @@ public void onPause() {
...
-			defaultTabPref.setSummary(String.format(getString(R.string.pref_summary_default_tab), defaultTabPref.getEntry()));
+		if(key != null) {
+			if (key.equals(getString(R.string.pref_key_default_tab))) {
+				// If the user changed the Default Tab Preference, update the summary to show the new default tab
+				ListPreference defaultTabPref = (ListPreference) findPreference(key);
+				defaultTabPref.setSummary(String.format(getString(R.string.pref_summary_default_tab), defaultTabPref.getEntry()));
+			} else if (key.equals(getString(R.string.pref_youtube_api_key))) {
+				// Validate the entered API Key when saved (and not empty), with a simple call to get the most popular video
+				final EditTextPreference youtubeAPIKeyPref = (EditTextPreference) findPreference(getString(R.string.pref_youtube_api_key));
+				if(!youtubeAPIKeyPref.getText().isEmpty()) {
```
ram-on: I'm getting a NullPointerException on Android 7.0 when I enter an invalid key such as "123"

```java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.isEmpty()' on a null object reference```

This can be easily resolved if we also check whether `youtubeAPIKeyPref.getText()` is null.

app/src/main/java/free/rm/skytube/gui/fragments/PreferencesFragment.java
ram-on: Also, if a user has previously entered an API Key and taps on "YouTube API Key", i would like that in that case, the old key is displayed in the text box.

PR #104, opened 2017-04-25
When the Subscriptions Feed is the default tab, show the refresh dial…

atomjack:
…og when the app first starts.


ram-on:
Hi Chris,

I did not merge this pull request due to:
* Duplicate functionality:  onFragmentSelected() (vs setFragmentSelected()) was already implemented in the parent class.
* Lack of Object Oriented code;  i.e. `if(videoGridFragmentsList.get(tab.getPosition()) instanceof SubscriptionsFeedFragment)` ...etc.

In any case, I've used your idea and implemented the fix (refer to 3166270540d39bac1811e002f28460aedc363059).

PR #110, opened 2017-05-12
Added (European) Portuguese translation

Generator:
Consider in the future adding skytube on a translation service like transifex, weblate, crowdin or other. 
For contributors easier access.

I've added an escape on "error_unable_to_subscribe", tell me if that's correct or if need to change
`Não é possível anular\/subscrever o canal (id=%s)`


ram-on:
Thanks

PR #125, opened 2017-06-23
Moved ChannelBrowserFragment back into MainActivity. Removed Fragment…

atomjack:
…HolderActivity.


ram-on:
Can you modify this PR such that the Channel Browser is displayed once a user clicks/taps on a user's avatar (when viewing the comments of a video)?

atomjack:
Hmmm I'm confused, do you mean when watching a video, dragging the comments in from the right, and tapping a commenter's icon? Because right now I noticed that doing that shows the replies to that comment. And, what channel would be displayed when clicking on a user?

ram-on:
When the user's avatar is clicked, the channel browser will be displayed for the user's channel.

atomjack:
Ok, done. I rebased and force pushed (had to fix some conflicts from my video options menu commit).

Some notes on the clicking of the user's icon in the comments list:

I had to make use of GetYouTubeChannelInfoTask (from YouTubePlayerFragment), so I moved it into its own class file, and added a YouTubeChannelInterface interface that it makes use of, to pass the YouTubeChannel object back.

CommentSnippet.getAuthorChannelId returns an Object, which is actually an ArrayMap (not sure why it doesn't just return an ArrayMap). While testing I came across an instance where this Object that was returned was null (not sure why, the resulting author name was blank as well), which caused the app to crash (when trying to get the YouTubeChannel from the null ChannelId), so CommentsAdapter checks to make sure the ChannelId is not null before getting the YouTubeChannel from it. If it is, clicking on the offending icon will just not do anything.

ram-on:
@atomjack Hi Chris, Apologies for the late reply.  I've tested your code and its seems that when:

1. I'm watching a video
2. Tap on the comments section
3. Click on a commenter's channel and his/hers channel is displayed (so far so good)
4. When I press the back button, the channel's fragment does not close ... in fact a white screen is displayed.  In order to go back to the video, I need to press the back button twice.

Can you look into it?

atomjack:
Hey @ram-on, somehow I totally missed your comment for all this time. I'm working on it now, and I found a fix. The problem is when we use `transaction.addToBackStack(null);` in `switchToFragment` - since we're calling MainActivity (and telling it to view a particular channel) from YouTubePlayerActivity (i.e. a separate activity from MainActivity), it's launching a new instance of MainActivity, which is then switching to the ChannelBrowserFragment, however you're not supposed to use addToBackStack for the first fragment you add. This is why the blank screen is showing up.

In order to get around this, I've had to add a boolean to MainActivity (called `dontAddToBackStack`) that gets set to true when an Intent to view a channel is received. The method to switch to another fragment will use this boolean to know whether or not to add to the back stack (and of course the boolean gets reset to false).

Right now the only way we're receiving this intent is from the YouTubePlayerFragment as well as the CommentsAdapter, which only appear in YouTubePlayerActivity, so for now it's safe to assume that that intent will only come from that other activity (instead of from MainActivity). The other ways we are switching to the ChannelBrowserFragment is via `MainActivityListener.onChannelClick`, so in those cases using `addToBackStack` is ok. One thing we could do is have the intent include a boolean that gets set to true for these two cases, in case we want to support receiving the intent from within MainActivity (if we don't, and we receive the intent from MainActivity, pressing the back button after the channel browser shows up will result in the app minimizing). Let me know if you think we should do that. Otherwise, if you think the changes look good, let me know and I will squash these commits so you can merge.

ram-on:
Hi Chris, I'm going to merge this.  In the meantime, pls add comments to YouTubeChannelInterface and its methods.

ram-on:
@atomjack I've just noted that this PR is causing the app to crash when I search for a video (using the search bar).  Can you look into it please?

atomjack:
@ram-on I just looked into this, and this branch is not the cause of the crash. It looks like it's because of the update to sdk 26. I found this: https://stackoverflow.com/questions/42961535/even-im-using-menuitemcompat-setonactionexpandlistener-my-app-crashes-with-advi, specifically the answer from ianhanniballake - possibly a bug in the 26.0.0-alpha1 release. 

I tested this by making new branches of both your "Android 8 - initial support" commit (which increased the sdk version to 26, including the version of the appcompat library), and the previous commit, "Added 'open with...' to gridviews videos menu", and the "android 8" branch crashes, while the other one doesn't.

PR #132, opened 2017-07-28
Update strings.xml

RemyL:


ram-on:
@RemyL Thanks for your translations.

For next time, please squash all commits into one.

PR #149, opened 2017-09-17
Added ability to export channel subscriptions from a Google Account, and import them into SkyTube. ...

atomjack:
...Moved Backups functionality into a custom class (that also contains the Subscriptions Import functionality) so that backups can be initiated from the Feeds tab (when there are no subscribed channels), as well as from the Preferences screen. Added Subscriptions Import to Backups Preferences Screen.

Some notes:

1. Since the backups import functionality needs to be accessible from the Feeds tab when there are no subscribed channels, I moved all that functionality out of `BackupPreferenceFragment` and into a new class, `SubscriptionsBackupsManager`, which also contains all the functionality for importing subscriptions from a Google/YouTube Account.
2. `SubscriptionsFeedFragment` is using the support library (v4) version of Fragments, while `BackupPreferenceFragment` (and the other PreferenceFragments) is using the native Fragment, and these two types of Fragments are incompatible with each other. Because of this, `SubscriptionsBackupManager` has two constructors - one using each type of Fragment. `SubscriptionsBackupsManager.hasAccessToExtStorage` will first look to see if the native Fragment was used, in order to request permission, and if not, it will look to see if the v4 Fragment is being used, in order to request permission. This seemed to me the easiest way since switching all the Preference Fragments to v4 would be a major pain.
3. Since you need to use either an Activity or a Fragment (either type) to receive the result of the permission request (whether or not they allowed permission), the Fragment (either type) that use the `SubscriptionsBackupsManager` class needs to pass this result on to the manager class, in `onRequestPermissionsResult`.


ram-on:
Hi Chris,

Tested it, looks great.

Just add comments to the following methods/classes:
* MainActivity#onResume()
* MainActivityListener#onSubscriptionsImported()
* ImportSubscriptionsAdapter
* ImportSubscriptionsChannel

atomjack:
@ram-on Got it, added the comments, and I squashed the commits and force pushed it.

PR #152, opened 2017-09-23
Updated gradle to version 3, and all included libraries to their late…

atomjack:
…st versions. Gradle update fixes crash when executing search.

I'll note that this requires using Android Studio 3 (currently Beta 6) - Android Studio 2 will not complete a gradle sync, it looks like it doesn't like gradle version 4.x.

The updates to the java files were due to using the latest version of Butterknife, which changes the annotation from @Bind to @BindView.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -28,18 +28,21 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
     }
+    flavorDimensions "skytube"
     productFlavors {
         // 100% Open-Source Edition
         oss {
             applicationId "free.rm.skytube.oss"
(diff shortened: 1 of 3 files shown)


ram-on:
I'm going to wait for Android Studio 3.0 to be released (soon hopefully).  I'll merge this PR once released.

ram-on:
@atomjack I've managed to fix the search crash by updateing the Android's support libraries to v26.1.0 while still being compatible with Android Studio 2.3 (af0b00d654bc981c4b2520d877cfbb05b3b6b5e4).  I've used the contents to this PR to update the other non-Google libraries.

I'm leaving this PR open and will merge it once Studio 3.0 is released.

ram-on:
Thanks!

PR #154, opened 2017-09-24
Spanish translation update

Freddynic159:
- Updated to the latest changes.
- Added and translated some missing strings.


ram-on:
@Freddynic159 Thank you for your contribution.

Regarding the country names, I've reverted them to English as country names should be an international thing rather than a localized one.

PR #155, opened 2017-09-24
Removed duplicate versions of English strings that were left in the S…

atomjack:
…panish Values file after Spanish translations for them were added.


ram-on:
Good catch 👍

PR #165, opened 2017-10-15
non-translatable resources

Generator:
Don't translate resources

Reference http://tools.android.com/recent/non-translatablestrings


ram-on:
@Generator Hello, I've reverted this commit as it was causing a lot of warnings:

https://i.imgur.com/2nofoJZ.png

Please fix those warnings.

PR #167, opened 2017-10-18
Ignore MissingTranslation

Generator:
This should fix the warning in https://github.com/ram-on/SkyTube/pull/165 .
Another approach is to move all non translatable strings to donottranslate.xml


ram-on:
@Generator It looks like those errors are no longer being displayed when using Android Studio 3.0.  I'm going to close this PR as it is no longer need.  Thx anyway!

PR #172, opened 2017-10-21
Upgrade

comradekingu:


ram-on:
Both Apple and MS appear to be using the term "Update Available" rather than "Upgrade Available".  Refer to http://www.google.com/images?q=update+available

comradekingu:
SkyTube uses upgrade elsewhere. Debian uses upgrade.

Issue #191, opened 2017-11-23
Enhancement: push notifications and auto-refresh


stemy2:
Hello,

To see new video, i'm forced to open the app and swipe down regularly. It would be nice to have a notification at every new video as in the official YT app.

Thanks.

atomjack:
I'm working on this right now. @ram-on, can I get your input on a few things?

1) It'll be using the Alarm Manager in order to trigger checking for new videos from subscribed channels. In Preferences -> Other, there will be a new option for "Feed Auto Notification", where you can select how often it should check for new videos from subscribed channels. The first option will be off, followed by 1 or more options, like every X hours. What do you think these options should be? 6, 12, 24? We probably don't want it happening too often, so as to not waste too many API requests.

2) When new videos are found, it will display a Notification. I was thinking, if only one new video happens to be found (out of all subscribed channels), the notification could display the Channel & Video title, along with a thumbnail for the video (and something like "New video from <Channel Name>: <Video Name>)". If more than one is found, a generic message saying "X new videos found from subscribed channels" (or something to that effect, I'm open to suggestions) instead.

3) When some hour option is selected (from whatever options we decide on), should it trigger a check for new videos right away? Or X hours (whatever option was selected) from now, for the first check (and then every X hours after that)? I'm thinking the latter. e.g. I select every 6 hours, so the next check will be in 6 hours from now, followed by another 6 from then, etc.

4) Supposing the latter, and the options are 6, 12, and 24 hours, let's say I set it to 6 hours, and some time later (a day, whatever), I go in to change it, and it happens that the next time it is supposed to check (at every 6 hours) is in 4 hours from now. If I change it to 12, should the next automatic check be in 12 hours from now, or 8 (12 - 4 = 8)?

Also, I don't think getting a notification for every new video will be practical, since we won't be able to use push notifications, and having it check too often will eat up the user's battery as well as API requests.

@ram-on, what do you think? @stemy2 I'd appreciate your thoughts as well.

ram-on:
Hello @atomjack 

1. I agree with 6, 12, 24 hours
2. I would keep it simple and always display a notification saying something like "X new videos uploaded from subscribed channels".  Once the user taps the notification, the app will open the feed tab as you suggested.
3. I prefer that such check is carried out X hours after the option is selected (in order to conserve the api quota)
4. If the user changes the settings, the check will be carried out X hours later.  I don't see that as a problem to be honest.

atomjack:
[...]

PR #194, opened 2017-12-01
Added Service to periodically check for new videos for Subscribed cha…

atomjack:
…nnels. For #191.


ram-on:
Hi @atomjack I've tested it out and changed the polling time to 1 minute.  However I've noted that the "Fetching Videos" dialog is not being displayed.

atomjack:
@ram-on hmm yeah, I didn't write it to have the dialog displayed by the service. I'm not quite sure if it should, though, since it's supposed to be doing it in the background. I suppose at the very least it should cause the Feed tab to refresh (to show the newly found videos), but should it also display the dialog? I'm just thinking since the user could be scrolling through the feed tab at the moment the service happens to finish, and all of a sudden the dialog shows without the user being ready for it. What do you think?

ram-on:
I see.  Sounds good.

PR #202, opened 2017-12-09
Upgrade, shortened update_dialog_msg

comradekingu:


ram-on:
I don't agree with these proposed changes.  Sorry mate.

comradekingu:
@ram-on The idea was to avoid the difference between apt-get update (pulling in new information) and apt-get upgrade (actually replacing programs).

PR #207, opened 2017-12-18
MainFragment should save the state of its Child Fragments, when the A…

atomjack:
…ctivity needs to be destroyed, and restore them when it's recreated. Not doing this was causing the child fragments to be recreated, and for the currently selected one to be set not selected, which was preventing the Refresh Dialog from showing when the default fragment was the Subscriptions Feed Fragment.

@ram-on It turns out that fix I implemented awhile back (https://github.com/ram-on/SkyTube/pull/169/commits/769bcd45dfbc4727880c6484a9d92c8dcf3d1483#diff-53cf7c2eb7dea2c5d7afeb3fb6c0b648R114) was incomplete. I'm not sure how I missed it. I figured out what the problem was by adding a bunch of logging, and when the activity gets recreated (again by going to Settings->Developer Options->Don't Keep Activities), I noticed that when I would log the `subscriptionsFeedFragment` instance in MainFragment, its identityHashCode was different from when I'd log `this` from within `SubscriptionsFeedFragment`. So, they weren't the actual same instance in memory (and `SubscriptionsFeedFragment` wasn't set as selected). 

To illustrate this, I created a branch (which doesn't have the fix this PR has) that adds the Logging I had done: https://github.com/atomjack/SkyTube/tree/feed-refresh-fix-new-logs. Another change that branch has is that instead of only refreshing every 3 hours, it refreshes every 10 seconds. Here's some steps you can take to see the problem (with the `feed-refresh-fix-new-logs` branch):

1. Set the option to not keep activities.
2. Fire up the app, wait for it to finish its initial refresh, make sure the Feed tab is set to be default (and selected), then go back to your home screen.
3. In Android Studio, filter the logcat output by "refresh" (all of the log lines I added have this word in them).
4. Wait at least 10 seconds, and then fire up the app again, which causes the activity to get recreated. 

If you look at the logcat lines, you'll notice the line that's being outputted from `SubscriptionsFeedFragment.GetTotalNumberOfChannelsTask`, as well as one from MainFragment:
```
12-18 13:27:04.704 5218-5218/free.rm.skytube.oss D/MainFragment: refresh stuff, setting current tab as selected: SubscriptionsFeedFragment{3661c5a}
12-18 13:27:04.892 5218-5218/free.rm.skytube.oss D/GetTotalNumberOfChannelsTask: Should refresh true, is fragment selected? false, this: SubscriptionsFeedFragment{5f65c44 #0 id=0x7f0900b8 android:switcher:2131296440:2}
```

You'll see that the identity hash code of `SubscriptionsFeedFragment` in `MainFragment` is `3661c5a`, yet inside `SubscriptionsFeedFragment.GetTotalNumberOfChannelsTask`, its `5f65c44`, and that instance of `SubscriptionsFeedFragment` wasn't set as selected. Once I implemented the change in this PR, of saving MainFragment's child fragments, the identity hash codes matched up, and the refresh dialog appeared as expected (because `subscriptionsFeedFragment` was set as selected).

You can ignore the `feed-refresh-fix-new-logs` branch after that, but if you try out this branch (`feed-refresh-fix-new`), edit line 79 of `SubscriptionsFeedFragment, and set it to look at a smaller time, like 10 seconds ago (`new DateTime().minusSeconds(10)`), and if you try the above steps again (ignoring the logcat parts), you'll see the refresh dialog show, as expected.

### DIFF
--- a/app/src/main/java/free/rm/skytube/gui/fragments/MainFragment.java
+++ b/app/src/main/java/free/rm/skytube/gui/fragments/MainFragment.java
@@ -50,6 +50,7 @@ public class MainFragment extends FragmentEx {
 	public static final String MOST_POPULAR_VIDEOS_FRAGMENT = "MainFragment.mostPopularVideosFragment";
 	public static final String SUBSCRIPTIONS_FEED_FRAGMENT = "MainFragment.subscriptionsFeedFragment";
 	public static final String BOOKMARKS_FRAGMENT = "MainFragment.bookmarksFragment";
+	public static final String DOWNLOADED_VIDEOS_FRAGMENT = "MainFragment.downloadedVideosFragment";
 
 	private VideosPagerAdapter			videosPagerAdapter = null;
 	private ViewPager					viewPager;
@@ -66,6 +67,7 @@ public class MainFragment extends FragmentEx {
 			mostPopularVideosFragment = (MostPopularVideosFragment) getChildFragmentManager().getFragment(savedInstanceState, MOST_POPULAR_VIDEOS_FRAGMENT);
 			subscriptionsFeedFragment = (SubscriptionsFeedFragment)getChildFragmentManager().getFragment(savedInstanceState, SUBSCRIPTIONS_FEED_FRAGMENT);
 			bookmarksFragment = (BookmarksFragment) getChildFragmentManager().getFragment(savedInstanceState, BOOKMARKS_FRAGMENT);
+			downloadedVideosFragment = (DownloadedVideosFragment) getChildFragmentManager().getFragment(savedInstanceState, DOWNLOADED_VIDEOS_FRAGMENT);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


ram-on:
Hi @atomjack.  Happy Christmas!  I've tested this and it does not seem to be working.  For example, I've added a logging operation whenever GetMostPopularVideos#init() is called.  This method is always called when the app is "minimized" and then "maximized"...

atomjack:
Merry Christmas @ram-on!! 

Hmm I don't think that has anything to do with this PR/branch - this PR just has to do with making sure whichever Fragment is selected when the activity is restored **knows** that it is the selected Fragment.

I switched to master, added logging to `GetMostPopularVideos#init` and `GetFeaturedVideos#init`, and they show up there as well, but only when I have "Don't Keep Activities" checked. But, since this appears in master as well, it doesn't have anything to do with this PR.

atomjack:
@ram-on I've updated this branch to include the new DownloadedVideosFragment, so it's ready to merge once we agree on a resolution to the issue you brought up. (I had to rebase master onto it, btw)

ram-on:
Happy new year @atomjack.

I've noted a bug that can be reproduced as follows:

1. Set Downloads tab as the default tab
2. Quit the app
3. Launch SkyTube
4. Minimize SkyTube -- it should now crash.

Exception:
`java.lang.IllegalStateException: Fragment FeaturedVideosFragment{87a433c} is not currently in the FragmentManager`

atomjack:
Happy New Year @ram-on! 

I've fixed that bug. It turns out I needed to set the offscreen page limit on the ViewPager that MainFragment uses to be the number of fragments used minus 1. Here's the stack overflow comment that clued me into it: https://stackoverflow.com/a/9857921/549602

An alternate solution to this is to check to make sure each fragment is added before saving the fragment (e.g. `if(featuredVideosFragment != null && featuredVideosFragment.isAdded())`), however I think setting the offscreen page limit to 4 should do for now, as I doubt we'll be adding any more tabs any time soon, and if we do, we might want to re-architect things a little for smaller devices (as you've probably seen, in portrait on a phone the tab names are abbreviated or plain cut off, with 5).

Issue #123, opened 2017-06-23
Feature request: Pre-downloading videos


gzsombor:
Sometimes I would like watch Youtube videos, when there is little, or no connectivity - it would be nice, if SkyTube could  
* download videos to show later - videos which are bookmarked, or marked as 'watch later'
* download new videos from a specific channel automatically

What do you think, this request is technically and legally feasible? I don't know, what are the rules using Youtube. Have you planned similar thing?

ram-on:
I'm planning to have the download video/music feature - this will cover your first point.

Watch later will be implemented however i don't think it will download the video automatically.

atomjack:
@ram-on I've just added a branch that fulfills this request. Would love to hear your thoughts. Here are my notes:

1) Each video cell (in the video grid) now has a download button, directly to the left of the video options button. Clicking the button will initiate the download, and the button will animate to show that the video is downloading (and then it will disappear once the download is finished).

2) I've added a 5th tab to the View Pager, for Downloads (and also added it to the default tab options). It works similarly to the Subscribed/Bookmarks feeds, but on a separate database for downloaded videos.

3) It uses the Android builtin DownloadManager to do the actual downloads. Each GridViewHolder will register a BroadcastReceiver (only upon download link click) to receive the Broadcast from the DownloadManager that the video is completed downloading. It then saves the URI to the local file in the DownloadedVideosTable (and unregisters the BroadcastReceiver).

4) I've implemented listeners for the DownloadedVideosDb, similar to how it's done for Bookmarks. All 5 fragments will be listeners, though, since any video can be downloaded/deleted, and if the video is on any of the other tabs, they should be refreshed to reflect the change (mainly showing/hiding the download button). However, the refresh won't happen for the currently selected Fragment, since the video's GridViewHolder will update itself to reflect the change (hiding/showing the download button).

5) I've moved the code that gets the desired stream for a Video into YouTubeVideo.java, since it's really only relevant for that class. It uses an interface to return the found stream. 

6) One thing I've noticed is that getting the desired stream can take a little while (more than a few seconds, sometimes), and of course only then can the download actually proceed. Not sure of a way around this.

I've tested this by downloading a video and then turning Airplane mode on, and it works :)

ram-on:
@atomjack I agree with this approach.  The only reservation I have is about the download button as I was more thinking about adding it into the three-dot-menu.

atomjack:
[...]

PR #208, opened 2017-12-19
Added feature to download videos for local/offline playback. Fulfills…

atomjack:
… #123.


atomjack:
Oh yeah, this PR shouldn't be merged until after I can update it after #207 is merged, so I can have the new `DownloadedVideosFragment`'s state saved along with the `MainFragment` child fragments.

ram-on:
Hello @atomjack.  I've noticed a bug which can be reproduced as follows:

1. Download a video that is not available in your country (e.g. https://www.youtube.com/watch?v=QIHhz0czVUA)
2. Download another video in which the download process is successful.
3. The first video (i.e. the failed one) will still be displayed in the Downloads section.

ram-on:
Also pls add a progress bar when the Downloads tab is loading the videos.

atomjack:
@ram-on Great catch!! It turns out I wasn't really using DownloadManager properly. It was causing the BroadcastReceiver for the unavailable video to receive the notification that the other download was complete, without making sure the notification was for *that* video. I've fixed that, and also made sure to unregister that BroadcastReceiver when it encounters a StreamError.

Also, I found a bug that was crashing the app when trying to play an unavailable video (via options menu->Enter video url).

About the progress bar, what exactly were you thinking - do you mean showing an entry for the videos that are currently downloading, in the Downloads tab? If not, can you elaborate?

ram-on:
@atomjack During the app initialization process, when the list of downloaded videos is being loaded, the Downloads Fragment should display the progress bar (just like what the other tabs are doing -- in order to retain consistency).

atomjack:
@ram-on Are you referring to the fact that when you first start up the app, if you switch to the Downloads tab right away (or have Downloads set as default), that the contents of the Downloads tab is empty for about 2 seconds, until the list is populated? 

If so, I'll note that the same thing happens with Bookmarks. So perhaps that tab should show the progress bar as well?

The thing is, for the other three tabs, the Feed tab shows the popup that shows details on all the channels it's retrieving videos for, while the Featured and Most Popular tabs show the spinner. I'm guessing you want the spinner (aka progress bar) to show up for Bookmarks and Downloads, then?

ram-on:
> Are you referring to the fact that when you first start up the app, if you switch to the Downloads tab right away (or have Downloads set as default), that the contents of the Downloads tab is empty for about 2 seconds, until the list is populated?

Indeed.  Just for the sake of consistency.

I've now just noticed that the Bookmarks does not have a progress bar.  Can you enable it?  It shouldn't take long as the "infrastructure" is already in place -- just call VideoGridAdapter#setProgressBar()

atomjack:
@ram-on Got it. Ok, so, it turns out it's going to be a little more tricky than just using `VideoGridAdapter.setProgressBar()`. Since the Bookmarks & Downloads Fragments use their own VideoGridAdapter (in order to support drag & drop reordering), setting the progress bar doesn't really work, as `VideosGridFragment.videoGridAdapter` ends up with a reference to the progress bar and hides it before the tab is done showing the videos, when either the Bookmarks or Downloads tab is set to default. In fact, it looks like it's actually causing two instances of `GetYouTubeVideosTask` to be being launched in this case (one by VideoGridAdapter, and one by BookmarksGridAdapter/DownloadedVideoGridAdapter).

My solution to this is to completely get rid of `BookmarksVideoGridAdapter` and `DownloadedVideoGridAdapter` (instead having the two fragments just use `videoGridAdapter`), and modify `VideoGridAdapter` to support drag & drop reordering. However, in order for the video grid to actually allow reordering, you need to call the ItemTouchHelper.callback stuff on the gridView (which is happening right now anyways, in Bookmarks & Downloads Fragments). I've chosen to add a new abstract subclass of `VideosGridFragment`, called `OrderableVideosGridFragment`, that does the setup required in order to support drag & drop reordering. Also, since VideoGridAdapter now needs to save the new order of videos to the database (in `onItemMove()`), it will have a reference to a Database (actually, a new interface that the Downloads & Bookmarks DB classes implement, with the `updateOrder(List<YouTubeVideo> videos)` method) so it can tell the database to update the order of the videos in it. The `OrderableVideosGridFragment` also has a reference to this database, that it passes on to `VideoGridAdapter`, and the Bookmarks & Downloads Fragments will set the appropriate database.

I wanted to run this all by you since it's a bit of a change, but I think it actually simplifies things a bit, and will make it easy enough to add other Fragments that use their own database, like Bookmarks & Downloads do. Plus, it looks like the two fragments (when set to default) end up displaying their contents a lot quicker, since it's only doing a single `GetYouTubeVideos` call for the Fragment. 

If this all sounds ok, let me know and I will push up the changes to this branch.

ram-on:
Hi Chris, I agree with the above proposal.  One thing that I may add:  would it be possible to extend VideoGridAdapter such that the subclass can support the drag and drop feature?

atomjack:
@ram-on Certainly, I just updated the branch with a new commit. I also rebased master onto it so you'll want to delete your local copy and then pull it back down.

ram-on:
Thanks Chris.

Issue #163, opened 2017-10-13
Keep history on search string


mcitew:
Maybe request that the search string be remembered, maybe maximum of 50 strings, if overflows it can push out the old and less popular search, or something scheme.

ram-on:
Will be implemented.  I'm closing this as it is duplicate to #57

mcitew:
Bookmark is not the same as keeping history of search string.

Keep history search string is to east re enter the new search or modifying the video search.

Bookmark is to keep a record of searched video. 

I don't see how the bookmark feature solves this issue.

atomjack:
I just added this feature, how's this look? You can also delete the entries with the X button.

![screenshot_20180103-132800](https://user-images.githubusercontent.com/114672/34540659-3800edac-f08a-11e7-9394-00296a0a22d7.png)

ram-on:
@atomjack Hey Chris, it looks great.  Just add an option in the preferences to not record search history.

atomjack:
@ram-on Cool, should that be off by default? As in, it will record search history by default?

ram-on:
> it will record search history by default?

Yes pls

atomjack:
Done. When you check the box to disable it, it will clear the Search History Database.

ram-on:
Implemented as per 8029882522cad25dfe6c09826899dfa481a5eca0

PR #218, opened 2018-01-03
Added Search History. Fulfills #163.

atomjack:


atomjack:
@ram-on I've rebased master onto this branch, so it can be merged cleanly. Make sure you delete your local first if you wanna test it out again, though.

PR #223, opened 2018-01-10
New menuitems on videos

gzsombor:
Subscribe to channel and open channel menu items added to the video listing.

What do you think?

### DIFF
--- a/app/src/extra/java/free/rm/skytube/gui/businessobjects/YouTubePlayer.java
+++ b/app/src/extra/java/free/rm/skytube/gui/businessobjects/YouTubePlayer.java
@@ -28,8 +28,11 @@ import android.util.Log;
 
 import free.rm.skytube.R;
 import free.rm.skytube.businessobjects.YouTube.POJOs.YouTubeAPIKey;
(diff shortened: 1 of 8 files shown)


Caboosium:
Pictures?

gzsombor:
It's not too big feature, just two new menu item : 
screenshot_1515939627

ram-on:
Hello @gzsombor That menu is getting really cluttered with information.

Refer to this menu:
screen shot 2018-01-14 at 10 20 06 pm

If you can somehow merge the "Share via", "bookmark" and the "download video" into a single row (just like the above screenshot), then I'll find no objection into adding the "subscribe" option.  The "Open Channel" option is useless as that can be achieved by simply tapping on the channel's name.

gzsombor:
Ohh, good to know. It's not too obvious, that the name and the next line is actually touchable. I thought, it wasn't, because of the lack of the area. Probably, my fingers are too wide ... :) Some hint would be nice for such a poor user like me, to discover this

ram-on:
@gzsombor I see.  That may be solved if we add the channel's avatar under the video's thumbnail similar to this:
mockup

gzsombor:
Looks nice !

PR #224, opened 2018-01-11
basic blocking feature added

okan35:
Hi Ramon, I added a basic blocking feature in which user can block the channels he/she does not want. To block a channel, user must click block channel on the video then refresh. After this, the user will not get any videos from that channel that is blocked. 

edit: I fixed the commits, now you have all the changes that I made. This should work without a problem now.


Caboosium:
Pictures? I would like to see what this looks like.

okan35:
@Caboosium 
it is here https://imgur.com/as3hRHm

app/src/main/java/free/rm/skytube/businessobjects/YouTube/Tasks/GetYouTubeVideosTask.java
```diff
@@ -115,7 +122,26 @@ protected void onPreExecute() {
 
 	@Override
 	protected void onPostExecute(List<YouTubeVideo> videosList) {
-		videoGridAdapter.appendList(videosList);
+
+		try {
+
+			ArrayList<YouTubeVideo> youTubeVideoList = new ArrayList<>();
```
ram-on: The code that checks if a channel is blocked or not should be move to `doInBackground()` as we must never block the main/UI thread as otherwise the app will stutter.

Also there is no need for the `blockedChannelsDb` instance variable.  Just remove it and place it in the `doInBackground()` or call `BlockedChannelsDb.getBlockedChannelsDb()` directly.

atomjack:
Also, I have a question - is it possible to unblock a channel?

ram-on:
> Also, I have a question - is it possible to unblock a channel?

Yeah that's actually a good point.  Users should be able to unblock the channels via the Preferences.

okan35:
For now there is not a way to unblock a channel that is why I call this a basic blocking feature. But I want to implement something like downloads or bookmarks,like a tab, so people can see which channels are blocked and they can unblock it from there. 

Until I implement feature to unblock it, it just blocks channel and user does not get any videos from the blocked channel.

ram-on:
> But I want to implement something like downloads or bookmarks,like a tab, so people can see which channels are blocked and they can unblock it from there.

I disagree - it would be better if we put it in the preferences as the main UI screen is already cluttered with tabs.

okan35:
That is also alright. I already started to check preferences for this unblocking feature. But as I said it might take some time.

ram-on:
> That is also alright. I already started to check preferences for this unblocking feature. But as I said it might take some time.

Thanks.  I suggest that you add a `PreferenceFragment` similar to, say, the backups preference -- there add an option for the users to "Manage blocked channels".  Once the user clicks that option, a dialog will appear and the user can untick which channels he wants to unblock -- the implementation is very similar to the Backup Preferences youtube channel importation (i.e. refer to that dialog).

okan35:
I got it. I will work on it. Hopefully it will not take long to do that.

okan35:
@ram-on I am finishing this implementation. I am almost there. I just could not find info about onClick of dialog of multiselectlistpreference. I am able to show channels that are blocked. But I somehow could not reach the onClick of dialog. So I wanted to ask you if you know how to reach the onClick method of the last dialog that shows blocked channels. I haven't found it yet. And I want those checkboxes to be true by default so when user clicks and makes them false the channel will be unblocked.

You can check it here on video: https://streamable.com/nbw2n

ram-on:
Hello @okan35.  That looks great.

Regarding your question, have your tried `onDialogClosed()`?

okan35:
Instead of dialog I am able to get the checked values' names which are channel names so I am planning on using them to unblock channels.

Trying something like this.

` multiSelectListPreference.setOnPreferenceChangeListener(new Preference.OnPreferenceChangeListener() {
            @Override
            public boolean onPreferenceChange(Preference preference, Object o) {

                String channelNameFromPreference = o.toString();
                Log.d("AA", "onPreferenceChange: " + channelNameFromPreference);
/*
               for (String channelName : blockedChannelsDb.getBlockedChannelsListName()){
                   if (channelNameFromPreference.equals(channelName)){
                       blockedChannelsDb.remove(channelNameFromPreference);

                       //Log.d("", "onPreferenceChange: " )
                   }
               }
             */
            if (blockedChannelsDb.getBlockedChannelsListName().contains(channelNameFromPreference)){
                blockedChannelsDb.remove(channelNameFromPreference);
            }
                return true;
            }
        });`

okan35:
@ram-on @atomjack Channel blocking and unblocking feature is now working.

https://streamable.com/ac4p0

ram-on:
@okan35 Can you modify the code's logic as follows:  when a channel is unblocked (i.e. via preferences), then the channel will never be shown again in the MultiSelectListPreference.

okan35:
That happens already but you need to go to main page and refresh. After that the unblocked channels are not showed in preferences.

okan35:
@ram-on that is why I put that toast.

okan35:
@ram-on Hi Ram-on did you have the opportunity to check the feature as I personally would like to have this on the following updates. 

This is another screenvideo for you to check the feature.

https://streamable.com/eyjro

atomjack:
I have some suggestions:

1. I don't like the use of the toast when you click the Blocked Channel preference - that's not really what toasts are for. I think if you put that text underneath the "Blocked Channel" preference instead, it would make more sense.
2. Maybe change the "OK" button in that popup to "Unblock"? It's not too clear what it does since it's just labeled "OK".

okan35:
@atomjack Those suggestions sound right. I will make necessary changes.

okan35:
@atomjack I made the first change but for the second change I am not sure how to do it. I am not sure if I have to make a custom dialog for that like in back up preference. Other than that I could not find a method or something to change the default dialog's "OK" to "Unblock".

ram-on:
Btw move your code from the master to a new branch (say "block-channels") as it facilitates testing from my side!

okan35:
Hi @ram-on Can you please check it now ?  I assume it should not have that error now. Also I created the new branch.
https://github.com/okan35/SkyTube/tree/channel-blocking

ram-on:
Hi @okan35.  I've just tested it -- looks great.  Please find below my feedback:

**Bug:**
1. Play a video > Press the menu button > Tap on 'Block Channel':  The channel is not blocked.

**Code Improvements:**
1. Catch exceptions and log them using:  `Logger.e(this, "<Describe what were you doing>", exception);` instead of `Log.d(TAG, "doInBackground: " + e.toString());`
1. Don't use `Log.d("", ...)`.  Rather use `Logger.d(this, ...);`
1. There are a lot of log that are not adding a value to the code (these were probably used during the development).  Some of them can be removed.
1. Remove any dead code.
1. Add comments describing what the code is doing:  e.g. add comments describing checks in place to detect any blocked channels inside `GetYouTubeVideosTask#doInBackground() `

**BlockedChannel Database:**
1. `BlockedChannelsDb#remove()`:  the code pertaining to DB delete is duplicate (inside that method)
1. `BlockedChannelsDb#getBlockedChannelsListId()` and `getBlockedChannelsListName()`:  `Youtube_Chabnel_Id` should not be hard-coded but rather as a variable
1. `BlockedChannelsDb#add()`:  I think you can remove the below code:
    `String a = "PRAGMA table_info(BlockedChannels);";
    Cursor b = blockedChannelsDb.getWritableDatabase().rawQuery(a,null);
    Log.d("PRAGMA", "add: " + DatabaseUtils.dumpCursorToString(b));`
1. `BlockedChannelsTable#COL_CHANNEL_ID`:  Change its value to "_id" (for the sake of consistency)
1. `BlockedChannelsTable#getCreateStatement()`:  the part pertaining to COL_ID should be changed to `COL_ID + " INTEGER PRIMARY KEY ASC, " +` (as currently the COL_ID is not being used to store data)

**BlockedChannel Strings:**
1. Please move them to a seperate String file (say "strings_blockedchannels")

**Others:**
1. With respect to `GetYouTubeVideosTask#doInBackground()`:  Transfer the `try {} catch{}` code inside the `if (!isCancelled()) {}`
1. `BlockedChannelsPreferenceFragment`:  Change the "OK" button's text to "Unblock" -- use:  `multiSelectListPreference.setPositiveButtonText(R.string. ...);`

okan35:
Hi @ram-on I just all the things you have requested. I hope I did not miss anything. I tested on my own it seems there is not an error or a problem and this time I pushed the changes to the branch I mentioned which is https://github.com/okan35/SkyTube/tree/channel-blocking

ram-on:
Closed at per PR #248

PR #226, opened 2018-01-14
Consistent change, permanent peace

comradekingu:
I seem to keep re-patching the use of "update" for what has historically been "upgrade", so I found a way to avoid it.
Wi-Fi as per https://en.wikipedia.org/wiki/Wi-Fi


ram-on:
Thanks

PR #227, opened 2018-01-15
Fix exception during startup

gzsombor:
During startup, sometimes, the putFragment is throwin an exception, because the fragment is not yet added. I haven't debugged it more, why it happens.


atomjack:
@gzsombor Did you by chance see #227 ? Your PR is the alternate solution I found to a slight bug, but the solution I found, of setting the offscreen page limit to the number of fragments/tabs minus 1, seems to obviate the need for making sure the fragment has been added. (The reason it's minus 1 is because that number doesn't include the actively selected fragment/tab, so added all up you end up with every fragment/tab being included).

gzsombor:
Sorry, but #227 is this issue :)

ram-on:
This commit shouldn't do any harm... just an additional check.

PR #229, opened 2018-01-16
Spelling: Channel information, - found

comradekingu:
Like "map info"
Alternatively, "information about channels"
Found was just redundant

### DIFF
--- a/app/src/main/res/values/strings_feed.xml
+++ b/app/src/main/res/values/strings_feed.xml
@@ -3,8 +3,8 @@
 	<string name="feed">Feed</string>
 
 	<string name="subscribing_to_channels">Subscribing to channels…</string>
-	<string name="fetching_subbed_channels_info">Fetching channels\'s information…</string>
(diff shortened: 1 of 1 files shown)


ram-on:
Hello @comradekingu.  Regarding "Fetching channels's information…" this actually should be "Fetching channels' information…".  My mistake.  I personally do not like "Fetching information about channels"... it is too long.  Unless you have another suggestion?

> Found was just redundant
I agree.

comradekingu:
@ram-on You can fetch channel information from one or multiple channels. Your above statement is unclear as to what you think about this.
If you describe exactly what happens, maybe I could come up with something better.

ram-on:
@comradekingu its aim is to inform the user that the app is currently fetching information about channel**s**.

comradekingu:
@ram-on It is "channel information" plural too. It is the _concept_ that is conveyed, not the subject of channel plural.

PR #238, opened 2018-02-01
Added swipe controls

avently:
- allowed to change brightness and volume via vertical swipe
- seek video is supported too via horizontal swipe
- brightness is saved across restarts in sharedPreferences and will be applied automatically on video load
- cool vector icons for all screens

Found time and made a normal PR with less code than before and adapted to current code base.

### DIFF
--- a/app/src/main/java/free/rm/skytube/gui/businessobjects/OnSwipeTouchListener.java
+++ b/app/src/main/java/free/rm/skytube/gui/businessobjects/OnSwipeTouchListener.java
@@ -107,6 +107,10 @@ public abstract class OnSwipeTouchListener implements View.OnTouchListener {
 	public void onSeekStart() {
 
 	}
+
+    public void onSeekEnd() {
+
+    }
 	public void onGestureDone(boolean notStart) {
 
 	}
@@ -142,7 +146,7 @@ public abstract class OnSwipeTouchListener implements View.OnTouchListener {
 	private final class GestureListener extends GestureDetector.SimpleOnGestureListener {
 
 		private static final int SWIPE_THRESHOLD = 50;
-		private Type type;
+		private Type type = Type.NONE;
 		private MotionEvent startEvent;
 
 		@Override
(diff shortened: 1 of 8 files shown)


ram-on:
Hi @avently, I'll test it out this weekend and I'll get back to you.

ildar:
@avently , great to see you implementing this essential feature!
Could you please note that the scale for the three should better be
exponential rather then linear? Maybe better fixing it now rather then
later getting to it again?
Thanks again!
P.S. maybe leaving brightness linear, but volume and seek definitely
exponential...

avently:
@ildar you mean you want to skip more than a minute via one swipe? For example,  first 10% of the screen width will skip 20 seconds,  20% will skip 1 minute, then 2 minutes, 5 minutes, etc?

atomjack:
Well done, @avently !! Looks great!!

ildar:
Yes, you got it right. As in vlc and such.

avently:
@ildar now it's a little bit cooler

ildar:
Awesome! :)

ram-on:
Hello @avently, thanks for your implementation - looks great.  The following is my feedback:

1. Methods inside OnSwipeTouchListener that do not have any code should be changed to abstract.
1. `OnSwipeTouchListener#viewRect()` should be changed to private and to abstract.  Also add comments on what it does.
1. What does `onGestureDone()` do?  Also the parameter notStart is not being used - so we should remove that.
1. Add (javadoc) comments (i.e. `/** ... */`) to each method and explain what each parameter does.
1. For `adjustBrightness()` and `adjustVolumeLevel()` inside YouTubePlayerFragment:  Pls add comments describing the major actions taking place.
1. `YouTubePlayerFragment#formatDuration()`:  This can be changed to private and pls remove the static keyword.

avently:
Hello @ram-on 
OnSwipeTouchListener#viewRect() is important. It can't be changed to private static. It returns actual frame of the view that overrides this method. I can't get actual rect from touchListener without it. 

Other is done.

ram-on:
Looks good. Thanks

PR #241, opened 2018-02-03
Fix npe

gzsombor:
Which happens, if channel information is tried to retrieve, when there is no network available.


ram-on:
Thanks

PR #255, opened 2018-02-24
Check internet connection

okan35:
@ram-on Now when the app starts there is a check for internet connection(not for a network mobile-wifi) and when there is no internet, blocked channels video list would get null pointer exception, that is also fixed.


okan35:
I messed up some commits as you see here.

PR #258, opened 2018-03-05
Force reloading the downloaded files

gzsombor:
... or after a download finished or removed. 
Otherwise the download fragment is not get updated, because GetBookmarksVideos always return null, until a refresh is called


ram-on:
Thanks.

PR #259, opened 2018-03-05
Do not throw an NPE during opening the main screen

gzsombor:
when there is no network available.
In that case the "videosList" is null.


okan35:
Hi @gzsombor I actually fixed it in connection check pull request sometime ago.

gzsombor:
Interesting, I saw it yesterday, when I tried to fix problems with downloads list is not refreshed

okan35:
It is because it has not been ready for merge so ram-on has not done it yet. There are just small fixes I should do in the commit but could not do it for some time. Hopefully will finish it.

gzsombor:
Ohh, I see :)

ram-on:
Thanks

PR #263, opened 2018-03-08
Check connection

okan35:
Hi @ram-on I threw all changes on my master and pulled from your master again which means I had a clean master by using this SO [answer](https://stackoverflow.com/questions/6284809/how-can-i-pull-from-remote-git-repository-and-override-the-changes-in-my-local-r) Then I made a new branch and made the necessary changes by touching nothing else, by the way I also deleted all those branches on both local and remote. 

But I still see old commits, I don't know how it is possible. So please let me know if this pull request is good.

By rhe way I did merge upstream master and did these changes. Now there is a problem, the back button click and pause don't stop a video and video still plays on background.I don't think this is because of my changes.


gzsombor:
The new video player control is a "bit" buggy, you should switch to the "legacy" one, then your issues will be fixed :)

app/src/main/java/free/rm/skytube/gui/activities/YouTubePlayerActivity.java
```diff
@@ -20,8 +20,12 @@
 import android.app.Activity;
 import android.content.pm.ActivityInfo;
 import android.os.Bundle;
+<<<<<<< HEAD
+import android.util.Log;
+=======
```
gzsombor: You merged something incorrectly

okan35: It can be because of old messed up commits and I tried to clean but it is still there. Maybe ram-on can fix it when he adds these changes.

PR #267, opened 2018-03-13
channels are sorted alphabetically

okan35:
Hi @ram-on I added a method to SubsAdapter to sort channels. I tried it by importing channels from youtube subscriptions, by subscribing and unsubscribing. It looks like it is working fine. 

By doing this in SubsAdapter, I did not need to do any sorting in `SubscriptionsDb` or `GetYoutubeChannelsTask`.

### DIFF
--- a/app/src/main/java/free/rm/skytube/gui/businessobjects/adapters/SubsAdapter.java
+++ b/app/src/main/java/free/rm/skytube/gui/businessobjects/adapters/SubsAdapter.java
@@ -18,6 +18,7 @@
 package free.rm.skytube.gui.businessobjects.adapters;
 
 import android.content.Context;
+import android.content.SharedPreferences;
 import android.support.v7.widget.RecyclerView;
 import android.util.Log;
 import android.view.LayoutInflater;
@@ -48,6 +49,8 @@ public class SubsAdapter extends RecyclerViewAdapterEx<YouTubeChannel, SubsAdapt
 
 	private static final String TAG = SubsAdapter.class.getSimpleName();
 	private static SubsAdapter subsAdapter = null;
+	private SharedPreferences sharedPreferences = getContext().getSharedPreferences("myPref",0);
+	private SharedPreferences.Editor editor = sharedPreferences.edit();
 	/**
 	 * Set to true if the users' subscriptions channels list has been fully retrieved and populated
 	 * by querying the local database and YouTube servers...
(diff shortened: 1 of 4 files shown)


ram-on:
Hi @okan35.  This is not working.  You need to move the `sortChannelsAlphabetically()` from the `SubsAdapter` to `GetSubscribedChannelsTask`.

Also add a preference options in the preferences for users to enable such functionality.

okan35:
@ram-on I don't understand how it is not working because I just tried again and it is working totally as it should. If I move move the `sortChannelsAlphabetically()` from the SubsAdapter to `GetSubscribedChannelsTask`, it works for importing channels from youtube xml file to subscripton drawer but it does not work when you subscribe to a channel.

But If I put it in `SubsAdapter#getSubsList()` it works when user subscribes and unsubscribes and imports channels from youtube. I can put a gif video if you want.

ram-on:
Hi @okan35.  My bad, it is indeed working!

Can you pls add an option in the Preferences for users to enable or disable this feature?

okan35:
Hopefully I will finish that part today, I already made the necessary changes in layout.

okan35:
@ram-on now you can check it. Although in my opinion this feature should be default as it looks much better when channels are in a order, so in my opinion there is no need for option to choose. But still did it, hope it is done :)

ram-on:
Hi @okan35 I've found 2 bugs:

**Bug 1:**
1. Subscribe to a number of channels
1. Enable the "Sort channels alphabetically" preference option
1. Go to the main activity and open the subscriptions list.  The channels are not sorted.

**Bug 2:**
1.  Ensure that the "Sort channels alphabetically" preference option is enabled.
2. Close the app
3. Reopen the app
4. The channels are still not sorted.

okan35:
@ram-on both are fixed. I tried what you did in both bugs, seems fixed to me. please let me know.

ram-on:
It's working great.  Thank

ram-on:
@okan35 I've just noticed this bug:
1. set the option to sort the channels
1. close the app
1. open the app
1. channels are not sorted.

Can you fix it?

okan35:
@ram-on I tried several times but I can't get that bug.

Trial 1: subscribe some channels - check sort channels - directly close the app - open the app. 

Channels are still sorted as it must be because the sorting is directly done under that fragment which is ensured by shared preferences.

Trial 2: same way but after checking sort channels option , I checked subscription drawer to see situation which was fine then closed the app, opened the app they were still sorted. 

By the way I did these tests after cleaning the data of the app.

ram-on:
@okan35 I've retested it on a clean build.  The above bug is reproducible - maybe you've forgot to commit some code?  I'm going to open an issue jut for the sake of keeping things in track.

PR #289, opened 2018-04-11
Escape single quotes before searching SearchHistory Database.

atomjack:
The app was crashing if you used a single quote when searching for videos, because it's searching the SearchHistory database but not escaping the single quote. This PR fixes that.

To reproduce on master, simple start typing the word `it's` into the search box and the app will crash.

I'll note I tried using `DatabaseUtils.sqlEscapeString` but it still crashed.


ram-on:
Good catch

PR #312, opened 2018-07-06
Add subscribers count in video description V2

Tvax:


ram-on:
Hi @Tvax, `GetChannelInfoTask` is not only retrieving the total number of subscribers but also banners, thumbnail, channel description...etc - you get the gist.  This method is expensive not only in terms of API usage but also in data usage...  I would suggest that `GetChannelInfoTask` would only return the channel' subscriptions counter.

Furthermore, there are cases in which the number of dislikes is not fully displayed:
screen shot 2018-07-15 at 3

Refer to the above screenshot:  there is a missing digit with respect to the dislikes.

Issue #260, opened 2018-03-05
Subscription DB is not initialized if the app is started without network


gzsombor:
Because SubscriptionsDb.getSubscribedChannels after asking the database, it asks the Youtube servers about more channel informations. This throws an exception, but that exception is catched, and an empty list is returned. In the end, the UI shows the warning about that there isn't any subscription yet - which is freaky a bit. The bigger problem is, that the user needs to force-stop the app, and restart it, after they connect to a network to get back their subscriptions / video list.
 I suggest to store all the channel informations locally for the subscribed channels, what do you think?

ram-on:
Good catch.  We can either cache the subscribed channels' info or add the subbed channels inside a SwipeRefreshLayout.

PR #317, opened 2018-07-25
Save the base informations about the subscribed channels inside db

gzsombor:
This is to address #260. Currently the same network checks are executed as before on initialization, but maybe, this can be skipped, if every channel has recent enough informations.


ram-on:
@gzsombor Apologies for the late reply.  This does not seem to be working.

What I did:
1. Run SkyTube with the network on.
1. Turned off the network and killed the app.
1. Re-opened the app.  The subscriptions are not visible (just like before).

Am I missing something?

gzsombor:
Yep, after my refactor, I forgot to test this thoroughly.  Fixed this issue !

PR #321, opened 2018-08-07
Reduce the number of playback status check, and reduce the repetition…

gzsombor:
… between the 2 player implementation


ram-on:
thanks

PR #355, opened 2018-09-17
Lazy loading of the videos on the tabs

gzsombor:
Instead of sending a network requests for each of the tabs during initialization, it's better to do it, just before the tab is needed - so API calls can be reduced.


ram-on:
Hi @gzsombor.  I like the concept of lazy loading - well done!
There is a small issue with the playlists:
1. Browse a channel
1. Tap on "Playlists" tab
1. Tap on a playlist.  Bug:  the playlist does not show any videos.

Can you pls take a look at the above bug?

gzsombor:
Good catch, fixed it! I haven't used that fragment before.

ram-on:
Thanks.

PR #365, opened 2018-09-28
Added Chromecast Support.

atomjack:
This is a pretty big update, which should hopefully not change how anything works in the OSS variant, since only the Extra variant will support Chromecast due to requiring Google Play Services. However, this update did require changing the architecture of some of the Activity classes.

The main big change is the addition of a new abstract class, BaseActivity. Any Activity that needs to support having the Cast button in the toolbar, and optionally showing the Chromecast Controllers (via SlidingUpPanelLayout), will need to extend this class. The only two classes that currently need this are MainActivity and YouTubePlayerActivity. The latter needs this support as when the user is watching a video on the device, they can hit the cast button, select a Chromecast, and the video they were currently watching will then start playing (resuming at the current position) on the selected Chromecast. 

The OSS & Extra variants each contain their own version of BaseActivity. 

The Extra variant's version contains all of the code needed in order to show the Cast button in the toolbar, as well as show the SlidingUpPanelLayout once a video starts playing, in order to control playback. It works pretty much the same as the Netflix app does it. The Panel contains two Fragments, `ChromecastMiniControllerFragment`, which is a small bar at the bottom of the screen, with video title & playback buttons, as well as `ChromecastControllerFragment`, which is a full screen fragment accessed via either tapping the Mini Controller or sliding it up. This Fragment contains the video's Thumbnail, position slider, playback buttons, video details and description.

The OSS variant's version of BaseActivity contains a bunch of mostly no-op methods, some of which will be called at certain times via MainActivity, but don't need to do anything since there is no Chromecast support.

It's important that the superclass of `MainActivity` and `YouTubePlayerActivity` are not changed, as doing this will break many things (in fact, the OSS version will not even compile if `MainActivity`'s superclass is changed, as it calls a few methods that are in `BaseActivity`).

Also, there is a special version of `main_activity_menu.xml` for the Extra variant, that shows the Cast button in the toolbar. The OSS variant uses the existing version of this file, but any changes that need to be made to it (addition of new options, etc) should also be done on the Extra variant's version. I have left a comment in the main version of this file to indicate this.

Another change is the need to refresh the VideosGridFragment at certain times, namely when you are playing a video on the Chromecast and stop it, in order to update the playback status indicator in the grid item of the video that was just played. The way this is handled when playing a video on the device is via `BaseVideosGridFragment.onResume`. This is never called when playing and then stopping a video on a Chromecast; the VideosGridFragment is still visible and usable while the video is playing.

@ram-on, please try out the OSS & Extra variants of this branch if you could, and let me know if you notice anything wrong, or if you have any questions about any of the changes this branch includes. Thanks!


ram-on:
@atomjack Thanks Chris.  I've just tested it very briefly, looks like it works great on devices without Chromecast.

The only issue i've noted is that the text "video blocker" is being displayed in the title bar instead of the shield icon:  Refer to:  https://i.imgur.com/7noAbKj.jpg

I've noted that some methods/classes are not documented.  I'll provide you with the list later on.

ram-on:
Hey Chris, I've haven't forgotten about this PR.  It will be included in the next release.

atomjack:
Awesome Ramon! Thanks, been meaning to ping you :)

rakshazi:
Hello! Any update on this great feature?

atomjack:
@rakshazi This is pretty much ready to go, it just needs to be merged in.

@gzsombor I noticed you've been merging a bunch of PR's. Have you had a chance to look at this yet? I just merged master into it and fixed the conflicts it ended up with (just a few).

Also, the Play Services Cast Framework library (which is required for Chromecast support) has a much newer version out since I last worked on this, however it requires migrating to [AndroidX](https://developer.android.com/jetpack/androidx/migrate). Do you want me to make a PR that does this (off of master, not this branch - I will then update this branch again)? It looks like it's mostly just changing the implementation paths in build.gradle and the corresponding paths in imports in various classes in the project. ButterKnife has a newer version out that uses AndroidX as well, so this will be required to use that newer version too.

gzsombor:
I've started looking into it, and definitely this should be the next thing, what I wanted to merge. Thanks for updating it! 
 In the last couple of days, I've looked into the AndroidX upgrade, but it seems, that in the end, we need a newer Android version for target, and - which seems to be a bigger problem - is that every time, when I upgrade the glide compiler, the build fails with NullPointerException.
 So I fear, this will be a bigger and separate task.

gzsombor:
I've looked through the code, and I think, I mostly understand, what's going on. Mostly :wink: But as I don't have Chromecast compatible device, I can't test it live.
 I've found a couple of smaller things, which I think, easy to fix (minor code ended up in the OSS version, what it's not needed), the little bigger things, what needs more consideration is the playback status handling. I see, that you tried to fix the video grids to update the view always, and I really like the generic idea to have multiple PlaybackStatusListener - what I don't see, where these listeners ever get unregistered. I'm not sure, this won't cause other problems somewhere else.

PR #370, opened 2018-10-08
Re-calculate the pretty timestamp after 5 minutes

gzsombor:
... and also fix the downloaded video publish date, which should be refreshed as well.


ram-on:
thanks

PR #371, opened 2018-10-08
README edit

TheTechnician27:
Replaced HTTP with HTTPS where possible; removed punctuation from bullet points where not a full sentence; added hyphens to third-party and open-source.


ram-on:
Thanks

Issue #392, opened 2018-11-03
Feature request: Preferred Resolution under Mobile Network


MarcoIeni:
Under `Preferences -> Video Player` there should be an entry like Preferred Resolution, but for Mobile Network, something like "Preferred resolution under mobile network".

MarcoIeni:
I am working on this

PR #393, opened 2018-11-04
Added preferred resolution on mobile network

MarcoIeni:
Solved issue #392.

This is my first contribution to this project and in general to an android app, so please double check :)

I added a TODO because I think that the method isConnectedToMobile() of the class MobileNetworkWarningDialog should be in a class not related to the GUI in order to be called in other parts of the code. If you tell me in which class to put this method I can solve it.

If the preferred resolution on mobile network is not set, the general preferred resolution is used also under mobile network.


ram-on:
Look good. Thanks

PR #422, opened 2019-03-21
Update README.md

zero77:
I think this is a major selling point for this app and should be mentioned.


gzsombor:
Thanks, @zero77 !

PR #450, opened 2019-07-09
Migrated project to AndroidX.

atomjack:
@gzsombor I saw your comment in my Chromecast branch about difficulties migrating to AndroidX, and tried it myself. I saw the same error you did about updating the minsdk version, however it had a [link](https://developer.android.com/studio/write/java8-support.html) that showed you can instead just add java 8 support by setting source and target compatibility to 1.8. From there, I was able to compile and install the app just fine, without changing the min sdk version.


gzsombor:
Awesome, thank you! I've somehow always triggered that NPE during build.

Could you rebase this commit on top of the latest master? It's seems to be interfering with the chromecast code

atomjack:
@gzsombor Sure thing! I've rebased this branch onto master.

gzsombor:
Why [this BackButtonActivity](https://github.com/ram-on/SkyTube/pull/450/files#diff-df1cb834cbc485fe7a856b9f18bce61a) is needed? I can't see any reference for this class, neither in the Java code, nor in XML, it wasn't there previously.

atomjack:
@gzsombor Hmm not sure, I think it must have existed in the project before, but was no longer needed, and was added by accident in this branch. I've removed it and force pushed the branch back up.

gzsombor:
Thank you, it looks flawless now :smile:

PR #466, opened 2019-07-20
Reversed color scheme in search suggestions. Show 10 most recent sear…

atomjack:
…ch terms when first opening search box.


app/src/main/java/free/rm/skytube/businessobjects/db/SearchHistoryDb.java
```diff
@@ -66,9 +101,20 @@ public void insertSearchText(String text) {
 			ContentValues values = new ContentValues();
 			values.put(SearchHistoryTable.COL_SEARCH_TEXT, text);
 			getWritableDatabase().insert(SearchHistoryTable.TABLE_NAME, null, values);
+		} else {
+			getWritableDatabase().execSQL(String.format("UPDATE %s SET %s = datetime('now','localtime') WHERE text = '%s'", SearchHistoryTable.TABLE_NAME, SearchHistoryTable.COL_SEARCH_DATE, SearchHistoryTable.COL_SEARCH_TEXT));
```
gzsombor: I think, the last parameter should be 'text', and not SearchHistoryTable.COL_SEARCH_TEXT.

Or better, just call the updateSearchTextTimestamp(text) function

atomjack: Oh, good catch. I think I had written that line before I had written the updateSearchTextTimestamp(text) method. I've updated it to use that instead.

gzsombor:
Thanks !

PR #485, opened 2019-08-05
Spelling: carges, - whitespaces

comradekingu:
As per https://github.com/ram-on/SkyTube/blob/6c03e5d048c86d77c528ea9484a0444a004d39a8/app/src/main/res/values/strings_youtube_player.xml#L14

Spotted by roptat at Weblate https://hosted.weblate.org/translate/skytube/downloads/fr/?checksum=5ad4e71edf305427


gzsombor:
Thanks for the patch!

PR #498, opened 2019-08-09
Added preference to set custom video download folder.

atomjack:


app/build.gradle
```diff
@@ -112,6 +112,7 @@ dependencies {
     }
     implementation 'com.mikepenz:actionitembadge:3.3.2@aar'
     implementation "com.mikepenz:iconics-core:3.0.3@aar"
+    implementation 'com.github.hedzr:android-file-chooser:v1.2.0-final'
```
gzsombor: There is already a file-picker library in this app - com.github.angads25:filepicker - any particular reason, why didn't you use it? I don't want to stick to that one, but I would rather see only one library for this problem

atomjack: Ahhh I didn't see the other one. I just started looking into implementing that instead, but I'm really not a fan of it, and it doesn't support switching between internal and external sd card. 

I'd like to look into switching to using this one I found rather than the other one, what do you think? I can add the changes to this branch.

gzsombor: It seems that https://github.com/hedzr/android-file-chooser is still maintained, opposed to angads25/filepicker, so maybe better to switch to this new one.

atomjack: Added a commit which removes angads25/filepicker, and uses the android-file-chooser one for Subscriptions and SkyTube settings imports.

gzsombor:
Awesome, thanks!

PR #499, opened 2019-08-10
Add scroll to top of list when clicking on current tab

Tvax:


gzsombor:
Thanks for the patch, interesting idea. I would never thought about clicking the current tab without any clue that it will do something, but this could be convenient. What do you think @atomjack ?

Tvax:
A lot of Android fediverse apps have these feature. Plus, if I remember correctly the Android YouTube app has it too.

atomjack:
Yeah that's great!!

gzsombor:
Thanks, I learn something new, everyday:)

PR #503, opened 2019-08-11
Added Preference to show/hide tabs in MainFragment.

atomjack:
After making changes, the app must be restarted. I tried having it update which tabs are shown without a restart, but it was causing issues, and I think most users won't be using this option very often, so it shouldn't be a big deal to restart the app after using it.

### DIFF
--- a/app/src/main/java/free/rm/skytube/gui/fragments/MainFragment.java
+++ b/app/src/main/java/free/rm/skytube/gui/fragments/MainFragment.java
@@ -3,28 +3,33 @@ package free.rm.skytube.gui.fragments;
 import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
-import androidx.annotation.Nullable;
-import com.google.android.material.tabs.TabLayout;
-import androidx.fragment.app.Fragment;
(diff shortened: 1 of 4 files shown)


gzsombor:
What about adding a new preference key, something like pref_key_default_tab_name, and only if it's empty, to get the old key, and do the preference upgrade? This way, the happy path would be simpler - get a preference, set the default tab name, done. The upgrade code is only touched in the rare case, when there is no 'new' preference configured.
 And could you please apply an 'Extract Local Variable' refactor, to the 'SkyTubeApp.getStringArray(R.array.tab_list_values)' and similar calls? It makes the code a bit harder to follow.

atomjack:
@gzsombor Can you elaborate on what you mean by 'Extract Local Variable' refactor? Not sure what you mean.

gzsombor:
Sorry for not being clear, I thought about instead of writing  `SkyTubeApp.getStringArray(R.array.tab_list_values)` lot's of times in the same function, introducing a new local variable:
`final String[] tabListValues = SkyTubeApp.getStringArray(R.array.tab_list_values)`
And using this everywhere

gzsombor:
Thanks for the patches!

PR #506, opened 2019-08-12
Use new pref key for default tab.

atomjack:
I had forgotten to change the pref key when setting the default tab to the new pref key, so updating the default tab wasn't working properly.


gzsombor:
Thanks!

PR #507, opened 2019-08-12
Don't show Feed refresh dialog when clicking on Feed tab when no chan…

atomjack:
…nels are subscribed to.

While testing hidden tabs I noticed this. If you clear storage and start the app up, and then click on the Feeds Tab, it shows the refresh dialog even though no channels are subscribed.


gzsombor:
Nice catch! Thanks

PR #510, opened 2019-08-12
Newpipe backend

gzsombor:
This is a **big** change, and it's not yet finished, and totally **not bug free**.
The goal is to use NewPipeExtractor as the backend API, to overcome the limitations of the free keys.

**What works:**
- [x]   Fetching new videos from channels
- [x]   Getting video details

**Missing features:**
- [ ] Search calls should use the new backend
- [x] Channel details should use the new backend

**Known bugs / limitations:**
- [ ]  NewPipeExtractor only returns the day of the video publication, but not the seconds, so it seems, that every video appears at midnight. To improve the experience, if the video is published on the same day, currently we are saving as 'published one minute ago'
- [x]  Video description is retrieved with html junks in it
- [ ]   Channel refresh is not multithreaded, so it's a bit slow


atomjack:
What about using 'Published today' instead of "Published one minute ago"? I figure that's less misleading.

gzsombor:
Yes, that's bit less misleading. but the other problem is that we currently sort the videos by this publish date, and if lot's of videos appears at the same milliseconds, it looks strange. Imagine, you open the app, refresh the channels, appears 10 new videos, at the top of the grid - all dated as '8 hours ago'. A couple of hours later, you refresh again - from the notification, you see, that there are 2 new videos - but you can't see any new at the top of the grid. It appears somewhere between, or after the 10 videos retrieved earlier.
 One solution could be to add a 'retrieval_timestamp' field to the database, and sort accordingly. This last minute hack was a much simpler solution, which I could came up quickly.

gzsombor:
As this won't brake any existing functionality (I really hope), can be switched on for the brave ones, and I using it for a couple of days, I think, this can be tested by the brave souls, who build their own SkyTube app! :)

atomjack:
Hell yeah! I'll test this and let you know how it goes!

atomjack:
FYI, this seems to break Chromecast playback, even when the new preference is unchecked. I'm looking into it.

gzsombor:
Ups, sorry !

atomjack:
Hmm I think I spoke too soon, it doesn't look like it has anything to do with the NewPipe backend stuff - I'm still getting errors (on the Chromecast) even when using an older version of the app. Trying to figure it out.

atomjack:
False alarm!! I tried several old commits but same thing, the video player on the chromecast kept throwing some weird error. I eventually rebooted the Chromecast, and all works fine now! So weird that it just started up as soon as I tried out the NewPipe backend - I was using it (via SkyTube Extra!) just before I synced the merge of this. 

I did switch to the NewPipe backend, and it's working fine! Does this completely bypass the need for API keys now (as long as the preference is checked)?

gzsombor:
Good to hear! :smile: 
The backend is not 100% complete, unfortunately, it only replaces the most commonly used functionality:
* checking new videos for a channel
* getting detailed information about a video
* getting detailed information about a channel

What missing:
* getting playlists from a channel
* searching 
* featured/most popular videos
* getting comments

I think, everything could be implemented with NewPipeExtractor, just the lack of time from me.
Hopefully, with this, the usage of the keys will be reduced in the future.

atomjack:
Cool. I was thinking of trying to tackle doing simultaneous channel videos fetches, to speed the subscription feed refresh up. Did you have plans to implement that or do you want me to work on it?

gzsombor:
Feel free to work on, anything, what you enjoy! 
Adding simultaneous channel video fetch could be a nice improvement indeed. I could write a long list of other improvements too :) Maybe we should start create individual tickets about improvement ideas, or create one Roadmap ticket... WDYT?

atomjack:
Cool, I've already started working on simultaneous channel videos fetches (for subscription refresh), will keep you posted.

But yeah, a roadmap ticket would be great! I also had an idea about doing audio-only streaming; having a button in the status bar that turns it on, so that it streams the audio track only. Could use a Service to support background playback (using ExoPlayer).

Also:
Playlists
Adding a Playlists tab so you can save favorite playlists, as well as creating your own?

Issue #475, opened 2019-08-01
Fc after the update


rex07:
08-01 19:28:27.880 E/m.skytube.extr(15140): Not starting debugger since process cannot load the jdwp agent.
08-01 19:28:33.327 E/AndroidRuntime(15140): Process: free.rm.skytube.extra, PID: 15140
08-01 19:28:33.327 E/AndroidRuntime(15140): java.lang.RuntimeException: Unable to start activity ComponentInfo{free.rm.skytube.extra/free.rm.skytube.gui.activities.MainActivity}: java.lang.RuntimeException: com.google.android.gms.dynamite.DynamiteModule$LoadingException: No acceptable module found. Local version is 0 and remote version is 0.
08-01 19:28:33.327 E/AndroidRuntime(15140): 	at free.rm.skytube.gui.activities.BaseActivity.onCreate(BaseActivity.java:122)
08-01 19:28:33.327 E/AndroidRuntime(15140): 	at free.rm.skytube.gui.activities.MainActivity.onCreate(MainActivity.java:106)

gzsombor:
What happened?

It seems to be coming from the Chromecast [code](https://github.com/ram-on/SkyTube/blob/a7a82e8fcbc7005a2123f4cb5b13f6957e66214b/app/src/extra/java/free/rm/skytube/gui/activities/BaseActivity.java#L122), but I can't reproduce it. @atomjack any idea, what could be the problem?

atomjack:
Hmm I did find [this](https://stackoverflow.com/questions/38251934/no-acceptable-module-found-on-android-emulators-using-com-google-android-gms-p). @rex07, do you see anything like this in your logcat?

```W/GooglePlayServicesUtil: Google Play services out of date.  Requires 9256000 but found 9080030```

Are you running it on a device or in an emulator?

rex07:
My device has no g apps installed maybe this is why the app fc

atomjack:
@rex07 Yeah, that's why it's crashing, you don't have google play services installed. I've never tested it on a device without gapps, so it got by me. Chromecast support requires google play services, although obviously it shouldn't crash if you don't have that installed. 

@rex07 can you try [this apk](https://drive.google.com/open?id=13y-9q_Cbb8mFf0_YtHGk7CqO0DjC3jo6), and let me know if it still crashes? 

[...]

PR #511, opened 2019-08-12
Make sure Google Play Services is installed before attempting to supp

atomjack:
…ort Chromecast. Show a popup just once if it is not installed (Extra variant only). Addresses #475.


atomjack:
@gzsombor If you'd like to test this out, create a virtual device that doesn't have Google Play Services installed, and install the app from the master branch, to see the crash. This branch fixes that.

gzsombor:
Thanks for the patch!

Issue #428, opened 2019-04-06
Preferred resolution for video download


austerenyan:
Add a setting to download videos in a specified resolution

atomjack:
@gzsombor this can be closed.

gzsombor:
Indeed! Thanks again :)

PR #515, opened 2019-08-14
Added Preference to select video download quality. Fulfills #428.

atomjack:
I chose to use a Preference as I thought it would be cumbersome to select the quality each and every time you try to download a video. 

From the couple videos I checked out, only 360p and 720p seem to be available in the StreamMetaDataList. However, I used all the available resolutions from the VideoResolution class.

Also, there's an option you can use for Android's DownloadManager to set a notification when a file is downloading and finished downloading, so I've added that to this branch.

### DIFF
--- a/app/src/main/java/free/rm/skytube/businessobjects/FileDownloader.java
+++ b/app/src/main/java/free/rm/skytube/businessobjects/FileDownloader.java
@@ -115,7 +115,8 @@ public abstract class FileDownloader implements Serializable, PermissionsActivit
 		DownloadManager.Request request = new DownloadManager.Request(remoteFileUri)
 				.setAllowedOverRoaming(allowedOverRoaming)
 				.setTitle(title)
-				.setDescription(description);
+				.setDescription(description)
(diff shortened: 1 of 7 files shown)


gzsombor:
Thanks for the patch!

gzsombor:
One thing, which I've noticed about the download notifications, is that when you click on them, it tries to open with anything, but the SkyTube app. Which is a bit unfortunate, I think.

atomjack:
@gzsombor When you download a video in SkyTube, and then later try to watch that video in SkyTube (e.g. via the Downloads tab), it streams it from the local copy, instead of from youtube's server. So, we could change it to show the notification during the download, but to remove it automatically when the download is finished, and show a Toast.

gzsombor:
Yes, that could work as well, but having a persistent notification is also a useful thing - sometime I've just click on a couple of videos, to download, and 10 minutes later when I return, it would be nice to know, if there were any failure, or everything is succeeded. Having a quick way to start downloads and see if a particular video is already available locally would be also nice - maybe with some simple overlay icons, what do you think?

PR #521, opened 2019-08-16
Added ability to bookmark videos by sharing youtube url's from other …

atomjack:
…apps. "SkyTube Bookmark" and "SkyTube Extra Bookmark" will now show up when sharing text from other apps.

Unfortunately it doesn't look like it's possible to restrict it to only show when sharing a string that is a valid video url, but trying to do so will simply show a Toast that only valid url's can be bookmarked.

I've tested this by sharing videos from the reddit app I use (Relay), and it works fine :)


gzsombor:
Thanks, nice idea!

PR #526, opened 2019-08-19
When using the NewPipe backend and fetching videos for subscribed cha…

atomjack:
…nnels, fetch up to 4 simultaneous channels at a time.

### DIFF
--- a/app/src/main/java/free/rm/skytube/businessobjects/YouTube/Tasks/GetBulkSubscriptionVideosTask.java
+++ b/app/src/main/java/free/rm/skytube/businessobjects/YouTube/Tasks/GetBulkSubscriptionVideosTask.java
@@ -17,28 +17,20 @@
 
 package free.rm.skytube.businessobjects.YouTube.Tasks;
 
-import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
-import java.util.Set;
-
-import org.schabi.newpipe.extractor.exceptions.ExtractionException;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import free.rm.skytube.businessobjects.AsyncTaskParallel;
-import free.rm.skytube.businessobjects.Logger;
 import free.rm.skytube.businessobjects.YouTube.POJOs.YouTubeChannel;
-import free.rm.skytube.businessobjects.YouTube.POJOs.YouTubeVideo;
-import free.rm.skytube.businessobjects.YouTube.VideoStream.NewPipeService;
 import free.rm.skytube.businessobjects.db.SubscriptionsDb;
-import free.rm.skytube.businessobjects.utils.Predicate;
 
 /**
  * A task that returns the videos of channel the user has subscribed too. Used
  * to detect if new videos have been published since last time the user used the
  * app.
  */
-public class GetBulkSubscriptionVideosTask extends AsyncTaskParallel<Void, GetBulkSubscriptionVideosTask.Progress, Boolean> {
+public class GetBulkSubscriptionVideosTask extends AsyncTaskParallel<Void, GetBulkSubscriptionVideosTask.Progress, Void> {
 
     static class Progress {
         YouTubeChannel dbChannel;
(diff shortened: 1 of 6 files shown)


atomjack:
I undid the change to use the NewPipe backend for subscribing to a channel as it caused a crash. Will look into it further.

atomjack:
Some notes after inspecting how this part of the NewPipe backend works:

There's a slight bug where, for channels that don't have any videos uploaded in the last month, it ends up grabbing the details for a bunch of that channel's videos, and reflects this in the popup's counter (of fetched videos), even though those videos aren't shown in the Feeds tab (or maybe they do, but they're so old you'd have to scroll down very far to find them). This also results in some channels taking a lot longer to fetch videos for compared to others.

One solution to this would be to upgrade NewPipeService.getUploadDate() to parse the dates that get returned from the initial fetch (that doesn't get the details). It seems like a lot of the videos I am seeing have a published date along the lines of "1 year ago", "2 months ago", "3 weeks ago", etc. I think if we modified that method to parse these into a usable date, we could use it to prevent the need to get the details for videos uploaded more than a month ago.

gzsombor:
Yes, I've seen this problem too. The publish date could be very tricky - because on my phone, I get the Hungarian equivalent of the '1 year ago' 'X month ago' strings. I guess, the default java net API adds Accept Language header based on the system settings. I think, we need to fix the HttpDownloader - in NewPipe, there is a much more complete implementation, which doesn't use the builtin java API, maybe that could work as well. 
 But long term, I think, parsing strings into dates, and handling special corner cases should be handled in NewPipeExtractor module, as this would benefit other client applications too.
 My other idea about, how to handle the infrequently updated channels - where there is one video in every months - involves 2 things:

1. Do not remove old videos from the subscription table (currently, they are removed if they are more than 1 month old)
2.  Implement paging on the subscription feed fragment - so having hundreds of videos in that tab wont be an issue, and later we can have a nice, offline filterable list of videos from subscribed channels.

What do you think?

atomjack:
I like those ideas. I'll see if I can come up with something for that. Let me know what you think of this branch once you have a chance to try it out!

app/src/main/java/free/rm/skytube/businessobjects/YouTube/Tasks/GetBulkSubscriptionVideosTask.java
```diff
@@ -51,43 +43,65 @@ public Progress(YouTubeChannel dbChannel, int newVideos) {
...
+                        }
+                        if(numTasksFinished < numTasksLeft) {
+                            if(tasks.size() > 0) {
+                                // More channels to fetch videos from
+                                tasks.get(0).executeInParallel();
+                                tasks.remove(0);
+                            }
+                            publishProgress(new Progress(channel, videos.size()));
+                        } else {
+                            videosDeleted = SubscriptionsDb.getSubscriptionsDb().trimSubscriptionVideos();
```
gzsombor: I would rather not delete any videos from the subscription db.

app/src/main/java/free/rm/skytube/businessobjects/YouTube/Tasks/GetBulkSubscriptionVideosTask.java
```diff
@@ -51,43 +43,65 @@ public Progress(YouTubeChannel dbChannel, int newVideos) {
...
+        numTasksLeft = tasks.size();
+
+        int numToStart = tasks.size() >= 4 ? 4 : tasks.size();
+
+        // Start fetching videos for up to 4 channels simultaneously.
+        for(int i=0;i<numToStart;i++) {
+            tasks.get(0).executeInParallel();
+            tasks.remove(0);
+        }
+
```
gzsombor: I know that the same code pattern is used in GetSubscriptionVideosTask - but this is too complex, hard to follow, and contains a couple of concurrency problems which could cause very strange behaviors - for example tasks execute multiple times, or tasks skipped entirely, or a locked UI (where the ui still thinks, that there are ongoing background tasks....)
 I would rather rewrite it with using proper java constructs - with a Semaphore and a CountDownLatch, with pseudo-code:

```
CountDownLatch countDown = new CountDownLatch(channels.size());
Semaphore semaphore = new Semaphore(4);
for (YouTubeChannel channel : channels) {
    semaphore.acquire();
    new GetBulkChannelVideosTask(countDown, semaphore).executeInParallel();
// in the GetBulkChannelVideosTask.doInBackground method:
///     ... actual network access...
///       semaphore.release();
///       countDown.countDown();
}

countDown.await();
// final result calculation..

}
```

gzsombor:
I've implemented a similar change, so I don't think this is needed now.

PR #534, opened 2019-08-26
When the app is resumed and the previous connection to a Chromecast f…

atomjack:
…ails (e.g. due to not being on wifi anymore), before the connection fails, attempting to play a video will wait until the connection to the Chromecast has been re-established. When this fails, the app should hide the spinner and play the video the device.

To elaborate, here are the steps to reproduce this bug which this branch fixes:
1) Start up the app, connect to wifi and connect to a Chromecast.
2) Kill the app (so it's not running anymore).
3) Disconnect from Wifi.
4) Fire the app up, and within the first few seconds of it opening, attempt to play a video.

The spinner will show, and just sit there spinning. This is caused by the fact that the app was previously connected to a Chromecast, so it attempts to reconnect. However, when this fails, the app keeps waiting for connection to be established, assuming it will eventually.

The new behavior will be that the spinner will show, but once the connection to the Chromecast fails, the spinner will be hidden, and the video will play on the device.


gzsombor:
Thanks, looks good!

PR #541, opened 2019-09-01
Russian translation

gzsombor:
Implement #527

Issue #574, opened 2019-09-29
[bug] "play video under mobile data" dialog can only be dismissed by double clicking "Play Video" 


tingsu:
- I have searched the history issue list, and have not found a similar bug.

- The bug was found in the latest release version 2.965, and were reproduciable on a Google Plxel 3 and an Android emulator 6.0 device

The environment: Only the mobile data is on; the wifi data is off.

- Bug: click one video, a dialog shows that warns whether to use mobile data (called it *Dialog 1*). click ``Play Video``, the video can be normally played (wait for a few seconds), exit the video, and click this video again from the main page, and this time, another dialog shows that asks whether to ``Resume`` the previous video or ``No``, Press ``No``, and *Dialog 1* appears, and click ``Play Video``, but this dialog cannot be dismissed, click ``Play Video`` again, this time the dialog disappears.

- Reproducing video

![SkyTube-Extra-2 965-issue-double-play-video](https://user-images.githubusercontent.com/3156978/65830053-5df37780-e2ac-11e9-8160-52c3842ed95f.gif)

gzsombor:
Strange, I can't reproduce it. For me, I can click on 'Cancel' or any other places on the screen (which means cancellation)

tingsu:
Hi @gzsombor , thanks for your reply.

On my side, I can always reproduce this bug. 

Note that this bug is **not** related to ``Cancel`` button. It is related to ``Play Video`` button. 

Let me state the reproducing step in more detail (The environment: Only the mobile data is on; the wifi data is off.)

Step 1: click one video, a dialog shows that warns whether to use mobile data (called it *Dialog 1*). click **``Play Video``**, the video can be normally played. This is correct behavior.

Step 2: Let the video to play for **more than 5 seconds** (this is **necessary** for reproducing this bug, this also lead to another issue #575 I reported), exit the video, and click this video again from the main page.

Step 3: At this time, another dialog shows that asks whether to ``Resume`` the previous video or ``No``, Press **``No``**, and *Dialog 1* appears, and click **``Play Video``**, but this dialog cannot be dismissed, click **``Play Video``** again, this time the dialog disappears.

The bug is deterministic. I can reproduce evey time on an Android 6.0 emulator.

This is how I create this emulator: 

``avdmanager create avd --force --name testAVD_Android6.0 --package 'system-images;android-23;google_apis;x86' --abi google_apis/x86 --sdcard 512M --device 'Nexus 7'``

tingsu:
[...]

PR #583, opened 2019-10-12
Do not open the same dialog twice, fixes #541

gzsombor:


gzsombor:
Obviously, this is for fixing #574

PR #595, opened 2019-11-01
Fix feed tab crash

Tvax:


gzsombor:
Thanks for the patch, it would solve the crash, however, it will introduce strange pagination errors - basically, after a video, which doesn't have a proper timestamp, the subscribed video list would reappear, from the newest again.
 After digging into the code, I found that the code doesn't use the proper columns from the database,  but get everything from the 'json' blob column - which makes everything much more complicated.

PR #640, opened 2020-01-22
Video playback speed controller added - ported the idea from NewPipe

gzsombor:


gzsombor:
@atomjack, what do you think, could this 'Playback speed' selector be improved from user experience standpoint? Now, it's super simple. And I've tried to improve the 'single-tap', 'double-tap' behaviour, so single-tap means to show/hide the controls, double tap is to play-pause the video, and after 5 seconds of play, the controls now auto-hided. Does this makes sense?

PR #667, opened 2020-02-25
Bump NewPipeExtractor to 0.18.4

gzsombor:


gillhash:
Does this effect Skytube!
 
NewPipe devs has updated Extractor to 0.18.5 with newer user agent forseeing youtube changes from March 2020.

https://support.google.com/youtube/thread/27596769?msgid=27596769

https://github.com/TeamNewPipe/NewPipeExtractor/releases/tag/v0.18.5

gzsombor:
Thanks @gillhash for the notification!

gillhash:
@gzsombor
Keep an Eagle eye on:
https://github.com/TeamNewPipe/NewPipeExtractor/issues/263

..as new user agent have broken some things !

Keep fallback option(to NP Extractor 0.18.4) Ready if needed.

wb9688:
Have you tried if this actually works with the UA you have in https://github.com/ram-on/SkyTube/blob/master/app/src/main/java/free/rm/skytube/businessobjects/YouTube/VideoStream/HttpDownloader.java#L50?

gzsombor:
Yes, indeed, the user agent needs that modification too.

Unfortunately, this 0.18.5 is a bit buggy. I hope, this will be sorted out quickly, and the March release will have a working backend in the near future

wb9688:
@gzsombor: Most issues should be fixed once I'm done with https://github.com/TeamNewPipe/NewPipeExtractor/pull/262. You could then use that commit hash instead of waiting until we release v0.18.6.

PR #673, opened 2020-03-01
Channel search feature for subscribed channels

okan35:
Hey guys, when I added sorting channels by name, I didn't implement the channel search. So I had some time today and made this in 20 min. To me it looks solid, Locale looks like working. In my opinion this is a crucial feature as when you have a lot of channels, it gets harder to scroll down and find the channel.


gzsombor:
Nice idea, I really like it!
Couple of small things:
- Why haven't you filtered on the db layer? I mean, passing the searchText into SubscriptionsDb.getSubscriptionsDb().getSubscribedChannelIds() and executing the query with an extra filter - something like this "lower("+SubscriptionsTable.COL_TITLE + ") like ?". 
- Checking for 'channel != null' should happen before the channel.title access, otherwise a NullPointerException will be the result. 
- I'm not a UI designer, but the search field with the white background looks a bit strange, why do you picked that color? The other search field is dark.

okan35:
I actually would like to do that but saw that this was a fast not really bad solution and also I am not %100 percent good on the codebase.

Yeah I saw that channel !=null after I pushed, will fix that.

The reason I did background white is because it is hard to see what you are searching for because of background color.

gzsombor:
I see, yes, for 20 minutes of work, this is a good feature :smile: 

What about having white fonts, and dark background? Just like the other controls...

okan35:
I mean sure we can do that too, it is just literally you don't see anything so either background or font color has to be different and plus there is this issue and not sure how to fix it as it could be from transparency of subs drawer, not sure. When you want to search for a channel you have to click first the search icon and then text view opens up for writing texts. If you just click on the search layout it will click to tabs behind subs drawer, you will understand better when you check yourself after I do those changes and push.

okan35:
@gzsombor made the changes. Also thought that we should not use this piece of code for sorting channels `// sort channels alphabetically (by channel name) if the user wants so...
			if (sortChannelsAlphabetically) {
				subbedChannelsList = sortChannelsAlphabetically(subbedChannelsList);
			}`
why dont we just sort channel at sql level ?

gzsombor:
Yes, you are absolutely right, that could happen on the sql layer too!

okan35:
I can also do that but not in this pull request. Another day when I have time,it won't take much time though.

gzsombor:
I think, it's good for merging, thanks for the idea, and for the patch!

PR #679, opened 2020-03-04
Sort channels by sql

okan35:
Just saw that there is a duplication problem in subs drawer after my chances.

 I will edit here when it is fixed.


gzsombor:
Thanks for the patch!

okan35:
@gzsombor I believe you didn't see my comment. There is a duplication problem with channels after this change. I will try to check this as it doesn't always happen or if you can check that also would be great as I haven't actually done a huge change in the process of sorting channels.

PR #684, opened 2020-03-09
Hide the Block Channel menu item when the Video Blocker is not enabled.

atomjack:
The video options popup menu should hide the 'Block Channel' button when the Video Blocker is not enabled.


gzsombor:
Thanks for the patch!

PR #687, opened 2020-03-14
Edit searched query

okan35:
It was annoying to rewrite a misspelled search text so I added this feature. User will be able to edit a searched text through a long press on searched text.


gzsombor:
Good idea! Thanks for the patch!

PR #693, opened 2020-03-18
Migrate PreferenceManager to androidx

ChemWater:
Migrated PreferenceManager class to android in the Settings class


gzsombor:
Thanks for the patch!

Is there any particular advantages to use androidx's PreferenceManager instead of the builtin one? I couldn't spot any differences ...

ChemWater:
No problem! I figured that the best way to ensure consistent behavior on all new devices would be to migrate all legacy code to androidx.

gzsombor:
Good point, thanks again !

PR #713, opened 2020-03-29
Improved backup

okan35:
Others preferences added to backup. On backup import all the preferences in others will be restored


gzsombor:
Why don't you put all the preferences in a map, and save it as a json? Using a database table, with a fixed structure will create issues, when new keys will added to the backup.

okan35:
I did this as a database because there is already existing db backup system, I basically put another db.getDbPath to zip method. 

I dont get what you mean by creating a json, do you mean I should have made json object parameter instead of different parameters for each preference ? And how would you backup json with db backups when you zip if you dont save it in a db ? if this is what you mean.

gzsombor:
You can zip and unzip any file, what you want, you don't need to convert it into SQLite db format.
Just create a JsonObject - or a Map - put the preference keys and values into it, and call Gson.toJson(preferences) and you will get a String, what you can write it into a file / or into the zip directly. For the other direction, you only need to call gson.fromJson(string).

okan35:
I wit try this then. I will literally make a json object and try to add into zip file we create in backup method.

okan35:
I tried somethings but I still had to make a file from json and then add in db, I could not do otherwise and on backup restore I do the same and get string from file but not really used gson on restore as I am not that knowledgable with gson. But this is what I got.
`{
   "pref_hide_tabs": [
      "MainFragment.featuredVideosFragment",
      "MainFragment.mostPopularVideosFragment"
   ],
   "pref_use_newpipe_backend": true,
   "SubscriptionsFeedFragment.FLAG_REFRESH_FEED_FROM_CACHE": false,
   "pref_key_screen_orientation": "auto",
   "pref_key_enable_immersive_mode": true,
   "pref_youtube_api_key": "",
   "pref_key_block_videos_lang_detection": false,
   "YouTubePlayerActivity.TutorialCompleted": true,
   "pref_key_channel_filter_method": "channel_blacklisting_filtering",
   "pref_key_subscriptions_alphabetical_order": true,
   "SkyTubeApp.KEY_SUBSCRIPTIONS_LAST_UPDATED": 1585931749830,
   "pref_key_brightness_level": 0.17374682,
   "pref_key_download_to_separate_directories": false,
   "pref_key_low_likes_filter": "-1",
   "pref_key_choose_player": "default",
   "Settings.LATEST_RELEASE_NOTES_DISPLAYED": "2.971",
   "SubscriptionsFeedFragment.FLAG_REFRESH_FEED_FULL": false,
   "pref_preferred_resolution_mobile": "5",
   "pref_key_enable_video_blocker": true,
   "pref_preferred_resolution": "5",
   "pref_key_low_views_filter": "-1",
   "pref_default_tab_name": "MainFragment.subscriptionsFeedFragment",
   "pref_key_preferred_region": "",
   "pref_key_mobile_network_usage_policy": "ask"
}`

gzsombor:
Thanks for the initial implementation, I could simplify a bit, and merged the code.

PR #714, opened 2020-03-29
Play back speed additions

okan35:
Added more playback speed options to play videos even faster  like a gif as sometimes personally I want videos to be played like a fast gif and there might be others who want this.

After speed 2x, I added like this 

`PLAYBACK_SPEEDS = {0.6f, 0.8f, 1f, 1.2f, 1.4f, 1.6f, 1.8f, 2f, 2.5f, 3f, 3.5f, 4f, 4.5f, 5f, 5.5f, 6f};`


gzsombor:
Not bad, this menu looks a bit strange - I don't know, what could be used instead of this - but it's usable.

PR #748, opened 2020-04-17
Latest AndroidX Migrations

ChemWater:
AndroidX migration for SkyTubeApp.java file


gzsombor:
Could you please rebase it? It seems to be based on an old version of the files.

gzsombor:
Rebased your changes in #770 - thanks for the patch!

PR #775, opened 2020-07-22
Added menu option to bookmark valid video urls when long pressed.

atomjack:
This branch adds a menu option when a valid YouTube video url is long pressed, to bookmark that video.

I chose to only have it attempt to bookmark the video, and not give the option to unbookmark an already bookmarked video, because in order to do that we have to create a YouTubeVideo object out of the url (in order to figure out if the video was already bookmarked or not), which takes a non-trivial amount of time. This would have to be done before the menu could be shown in order to show either "Bookmark" or "Unbookmark" as the new menu option. The result was that long pressing on a video url would introduce a delay of a few seconds before the menu would show up, since it would need to make a network call to construct the YouTubeVideo object.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: android
 dist: trusty
 android:
   components:
-    - build-tools-28.0.3
-    - android-28
+    - build-tools-29.0.2
+    - android-29
 env:
   - TARGET=assembleOss
   - TARGET=assembleExtra
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: android
 dist: trusty
 android:
   components:
-    - build-tools-28.0.3
-    - android-28
+    - build-tools-30.0.1
+    - android-30
 env:
   - TARGET=assembleOss
   - TARGET=assembleExtra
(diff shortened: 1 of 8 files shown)


atomjack:
This branch updates to the latest build tools but Travis CI is failing because of it. https://travis-ci.community/t/unable-to-specify-build-tools-30-0-1-and-accept-license/9389

gzsombor:
Any particular reason, why you needed the SDK update? I could be wrong, but I don't think this is needed for this feature.
I suggest to remove it, until we can't figure out, how to run the CI builds

atomjack:
@gzsombor Android Studio did it automatically. I'll remove those changes and push as soon as I'm able.

atomjack:
@gzsombor I changed it to 29.0.2 and that passes the checks. However, I think I can get it to show bookmark/unbookmark without having to construct a YouTubeVideo object and a network call. I'll post again soon.

atomjack:
@gzsombor Ok! Please give this a look.

gzsombor:
Looks good, I guess, we don't really need the regexp, because the id is already inside the ContentId (I know, the various id-types are a bit much, it's evolved organically, so to say :wink: ), but we can fix it later.

PR #794, opened 2020-08-08
Unsubscribe all channels change

okan35:
I changed unsubscribing by iteration to an sql statement by using "IN". Also removed handler and removal of channels from subsAdapter as it already was not doing anything, what we do in the unsubscribe method is sufficient enough to remove channels from subsadapter. 

I believe this should be more performant and is a cleaner way of doing it, from my tests things looked ok.


gzsombor:
After thinking about a bit, why don't we simply delete all the channels and videos? We don't need to get all the subscribed channel ids, and issue a long SQL to delete everything. 
getWritableDatabase().delete(SubscriptionsVideosTable.TABLE_NAME) and getWritableDatabase().delete(SubscriptionsTable.TABLE_NAME) could be enough, couldn't be?

okan35:
Actually you are very right, I totally didn't think of it, as we are unsubscribing user from all the channels so it doesn't matter if we just truncate the table as well.  I will do this and make a push later.

gzsombor:
Thanks for the improvement!

PR #798, opened 2020-08-15
Fragment error fix

okan35:
I got `Fragment <fragmentName> is not currently in the FragmentManager` error and the app crashed when I played a video.
This was the suggested fix, so I applied and so this does not go unnoticed.

Also if user can't get any videos then when user needs to be notified the uninitialized context gives null error and app crashes, I put a check there so it will not crash app for now and will put a log for us on logcat.

Also just improved isBoorkmarked and isSubscribed from 2 ms to 1 ms of time, nothing significant but just did it as I had time.


gzsombor:
Thanks for the patch, sorry, I've missed totally.

PR #843, opened 2020-12-03
Use Collection.removeIf().

Isira-Seneviratne:


gzsombor:
Thanks for the patch!

PR #846, opened 2020-12-07
Simplify showNotification().

Isira-Seneviratne:


app/src/main/java/free/rm/skytube/gui/fragments/SubscriptionsFeedFragment.java
```diff
@@ -193,38 +194,30 @@ public static boolean isFlagSet(String flag) {
...
+			PendingIntent pendingIntent = PendingIntent.getActivity(requireContext(),
+					1, new Intent(), PendingIntent.FLAG_UPDATE_CURRENT);
+			builder.setPriority(NotificationCompat.FLAG_ONGOING_EVENT)
 					.setContentIntent(pendingIntent);
+		}
 
-
-			NotificationManager notificationManager = (NotificationManager) getContext().getSystemService(Context.NOTIFICATION_SERVICE);
-			notificationManager.notify(NOTIFICATION_ID, builder.build());
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
gzsombor, line 212: Any particular reason to use a separate if, instead of a plain simple 'else' ?

Isira-Seneviratne: My bad, I'll change it.

gzsombor:
Thanks for the patch!

PR #847, opened 2020-12-08
Use Objects methods.

Isira-Seneviratne:
Enabling Java 8 support allows the `Objects` methods to be desugared when the `minSdkVersion` is below 19.

Screenshot 2020-12-08 064326


gzsombor:
Thanks for the patch!

PR #848, opened 2020-12-08
Use RxJava 3 in database tasks.

Isira-Seneviratne:
Replace the database `AsyncTask` implementations with methods that use RxJava 3, as `AsyncTask` is deprecated starting with API level 30.


gzsombor:
Interesting idea, and it doesn't look that scary as I originally thought, that introducing reactiveness would look :) 
I have only minor concerns around these CompositeDisposable objects - if I understand it correctly, with this change - as I don't see any places where we remove an already finished 'Flow' from a disposable, that CompositeDisposable will hold a reference to the 'flow'? Which, I fear, could introduce new memory leaks - I know, the current app has already have a couple of it, already :wink:  And similarly, putting that CompositeDisposable inside the YoutubeChannel is not the best place I believe, can we move that to a UI class?

app/src/main/java/free/rm/skytube/businessobjects/YouTube/POJOs/YouTubeChannel.java
```diff
@@ -247,7 +249,8 @@ public boolean blockChannel(boolean displayToastMessage) {
 
 		// if user is subscribed to the channel, then unsubscribe first...
 		if (SubscriptionsDb.getSubscriptionsDb().isUserSubscribedToChannel(getId())) {
```
gzsombor, line 251: I think, this 'isUserSubscribedToChannel' is also a db call, which could block.

Isira-Seneviratne: In that case, it could be converted to a `Single`.

gzsombor:
Thanks for your hard work!

PR #849, opened 2020-12-08
Use AndroidX Preference.

Isira-Seneviratne:


gzsombor:
Looks nice, I've found only one problem, previously, if you clicked on the back arrow on the main preference page, we went back to the MainActivity, now, nothing happens.
And a couple of questions:
1, Why the special handling of the VideoBlockerPreferenceFragment ? Is it because we want to directly navigate to that fragment? 
2, Is there any changes in AndroidManifest.xml, or it was just an un-intentional whitespace/reformat change, probably by IntelliJ?

Isira-Seneviratne:
> Looks nice, I've found only one problem, previously, if you clicked on the back arrow on the main preference page, we went back to the MainActivity, now, nothing happens.
> And a couple of questions:
> 1, Why the special handling of the VideoBlockerPreferenceFragment ? Is it because we want to directly navigate to that fragment?

Yeah.

> 2, Is there any changes in AndroidManifest.xml, or it was just an un-intentional whitespace/reformat change, probably by IntelliJ?

Yeah, I created the settings activity using Android Studio's activity creation wizard, which caused it to update the manifest.

gzsombor:
I've just finished testing your other change, and merged your code - so this will have merge conflicts :(

Isira-Seneviratne:
> I've just finished testing your other change, and merged your code - so this will have merge conflicts :(

I fixed those.

gzsombor:
Thanks for the patch!

PR #850, opened 2020-12-12
Enable code shrinking.

Isira-Seneviratne:
Code shrinking removes any unused code from the app, thus reducing the number of dex files the app contains.


gzsombor:
Thanks for the patch!

PR #851, opened 2020-12-12
Use try-with-resources.

Isira-Seneviratne:
Enabling Java 8 support allows [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) to be used on API levels below 19.


gzsombor:
Thanks for the cleanup!

PR #853, opened 2020-12-13
Rewrite the YouTube AsyncTask implementations as methods that use RxJava.

Isira-Seneviratne:


gzsombor:
Are you aware of the build failure? It's  not just my laziness which prevents reviewing and merging this huge changeset :wink:

Isira-Seneviratne:
Apparently, the compiler identified [`YouTubePlayer.launch()`](https://github.com/ram-on/SkyTube/pull/853/files#diff-edf6af3e95f3835152f650288cbc13b5cefd3ce34e1b5e205e2eb99cc227dbb6R91) as returning `void`, but it returns a `Disposable`.

gzsombor:
Unfortunately, the compiler knows better: https://github.com/ram-on/SkyTube/blob/0a0eaab29f8382a6aa2abe783f624aa3b89b44a0/app/src/oss/java/free/rm/skytube/gui/businessobjects/YouTubePlayer.java#L52

Isira-Seneviratne:
> Unfortunately, the compiler knows better:
> 
> https://github.com/ram-on/SkyTube/blob/0a0eaab29f8382a6aa2abe783f624aa3b89b44a0/app/src/oss/java/free/rm/skytube/gui/businessobjects/YouTubePlayer.java#L52

Got it, thanks!

gzsombor:
Unfortunately, I couldn't finish reviewing your patch, I'm trying to understand, how the various RxJava methods should be used in the app. I mean, doOnSuccess/blockingSubscribe/subscribe(...) looks very similar, and I'm not sure, that always the correct one is used here. Could you share, based on what, did you select one over other?
 The biggest and most complex background task is the bulk subscription fetcher, so I'm concentrated there, my questions:
* previously, it was multi-threaded, so multiple channels are fetched parallel - now, with that concatMap(...) construct, this is lost, isn't it? 
* I think, the 500 ms delay of updating the UI after subscription fetching, in the current implementation, is more of a hack, and not really needed - for me it looks a quick way to inject a task to update the UI.
* Ideally, every channel should be checked concurrently (4 or 5 at the same time), and when new videos were found,  'fetching the details' needs to be called on them. **And** the UI should be updated every 10 or 20 seconds. Currently, it is updated only at the end of the process, which is not the greatest user experience. I hope, with RxJava these kind of changes will be easy to implement.

Isira-Seneviratne:
> Unfortunately, I couldn't finish reviewing your patch, I'm trying to understand, how the various RxJava methods should be used in the app. I mean, doOnSuccess/blockingSubscribe/subscribe(...) looks very similar, and I'm not sure, that always the correct one is used here. Could you share, based on what, did you select one over other?
> The biggest and most complex background task is the bulk subscription fetcher, so I'm concentrated there, my questions:
> 
> * previously, it was multi-threaded, so multiple channels are fetched parallel - now, with that concatMap(...) construct, this is lost, isn't it?

The channels were being fetched in a for-loop. If you want, I could use the `flatMap` operator instead.

> * I think, the 500 ms delay of updating the UI after subscription fetching, in the current implementation, is more of a hack, and not really needed - for me it looks a quick way to inject a task to update the UI.

I see. I was maintaining the existing behavior.

> * Ideally, every channel should be checked concurrently (4 or 5 at the same time), and when new videos were found,  'fetching the details' needs to be called on them. **And** the UI should be updated every 10 or 20 seconds. Currently, it is updated only at the end of the process, which is not the greatest user experience. I hope, with RxJava these kind of changes will be easy to implement.

RxJava's types have plenty of useful operators: http://reactivex.io/RxJava/3.x/javadoc/

gzsombor:
> > Unfortunately, I couldn't finish reviewing your patch, I'm trying to understand, how the various RxJava methods should be used in the app. I mean, doOnSuccess/blockingSubscribe/subscribe(...) looks very similar, and I'm not sure, that always the correct one is used here. Could you share, based on what, did you select one over other?
> > The biggest and most complex background task is the bulk subscription fetcher, so I'm concentrated there, my questions:
> > 
> > * previously, it was multi-threaded, so multiple channels are fetched parallel - now, with that concatMap(...) construct, this is lost, isn't it?
> 
> The channels were being fetched in a for-loop. If you want, I could use the `flatMap` operator instead.

The current implementation use a for loop, but it put the individual network request in a thread pool - with a semaphore to limit the concurrency to 4 parallel threads, here:
https://github.com/ram-on/SkyTube/blob/7789a8beb05bddf458d57776e50363a57fbfeb91/app/src/main/java/free/rm/skytube/businessobjects/YouTube/Tasks/GetBulkSubscriptionVideosTask.java#L81

> 
> > * I think, the 500 ms delay of updating the UI after subscription fetching, in the current implementation, is more of a hack, and not really needed - for me it looks a quick way to inject a task to update the UI.
> 
> I see. I was maintaining the existing behavior.
> 
> > * Ideally, every channel should be checked concurrently (4 or 5 at the same time), and when new videos were found,  'fetching the details' needs to be called on them. **And** the UI should be updated every 10 or 20 seconds. Currently, it is updated only at the end of the process, which is not the greatest user experience. I hope, with RxJava these kind of changes will be easy to implement.
> 
> RxJava's types have plenty of useful operators: http://reactivex.io/RxJava/3.x/javadoc/

Thanks!

gzsombor:
Thanks for the patch! Unfortunately it became conflicting with other recently merged changes. I could rebase it, and fix a couple of smaller issues around threading in #894

gzsombor:
Sorry for the slow responses/review, but it's a big change, and I expect this triggers a couple more bugs

PR #857, opened 2020-12-14
Use view binding.

Isira-Seneviratne:


gzsombor:
Nice! Always a good feeling, when one can remove a lot of unnecessary code :+1:

gzsombor:
And this one also fails...

Isira-Seneviratne:
The `extra` variant builds fine, though.

gzsombor:
I see, yes, you can't refer to the chromecast related fields in the OSS variant.

gzsombor:
@Isira-Seneviratne , sorry, but the CI build fails and there is a lot of merge conflict in these patch, could you have a look on it? Do you plan to fix it?

gzsombor:
Thanks @Isira-Seneviratne  for the initial patch, I had to rebase and fix the compilation errors, but now it looks and works great!

PR #858, opened 2020-12-14
Use vector drawables.

Isira-Seneviratne:
Replace PNGs with vector drawables where possible.

**Screenshots**

Screenshot_20201215-051521491
Screenshot_20201215-051536122


gzsombor:
Nice, thanks for the patch!

I've thought about doing the same, but replacing the icons from https://github.com/mikepenz/Android-Iconics . What do you think?

Isira-Seneviratne:
Yeah, sounds like a good idea.

PR #861, opened 2020-12-20
Update prettytime to 5.0.0.Final.

Isira-Seneviratne:


gzsombor:
Thanks for the patch!

PR #862, opened 2020-12-21
Check if the current network is metered.

Isira-Seneviratne:
Check whether or not the current network is metered instead of checking if it is a mobile network.

### DIFF
--- a/app/src/main/java/free/rm/skytube/app/Settings.java
+++ b/app/src/main/java/free/rm/skytube/app/Settings.java
@@ -80,9 +80,9 @@ public class Settings {
         return getPreference(R.string.pref_key_hide_tabs, Collections.emptySet());
     }
 
-    public Policy getWarningMobilePolicy() {
+    public Policy getWarningMeteredPolicy() {
         String currentValue = getSharedPreferences().getString(getStr(R.string.pref_key_mobile_network_usage_policy),
-                getStr(R.string.pref_mobile_network_usage_value_ask));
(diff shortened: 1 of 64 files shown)


gzsombor:
Thanks for the patch. I only have two concerns:
* As the translations are handled by weblate ( https://hosted.weblate.org/git/skytube/strings/ ), if we just rename the keys/values no translator would notice, that the original text have been changed. Maybe, we can introduce new keys/translations, and use a fallback, if no translation found. But I think, I'm a bit over complicating our life's. :thinking: 
* Adding migration to this settings could be nice: https://github.com/ram-on/SkyTube/blob/f3e2392f63d663af03fe59a42af5804f945aa276/app/src/main/java/free/rm/skytube/app/Settings.java#L50

Isira-Seneviratne:
> Thanks for the patch. I only have two concerns:
> 
> * As the translations are handled by weblate ( https://hosted.weblate.org/git/skytube/strings/ ), if we just rename the keys/values no translator would notice, that the original text have been changed. Maybe, we can introduce new keys/translations, and use a fallback, if no translation found. But I think, I'm a bit over complicating our life's. 🤔

In that case, shall I revert the key name changes?

gzsombor:
I think, the simplest thing is to only modify the English text and the keys, and the translator can freely decide, what is the appropriate translations.

gzsombor:
Thanks for the patch,  I've rebased and merged it!

Issue #838, opened 2020-11-21
Ability to import subscriptions via a .json file 


vanillaneutrino:
Hi, is there a way of importing subscriptions via a .json from google takeout? The original method of exporting subscriptions from youtube does not work anymore. Thanks
Thanks.

profucius:
Second this. Google changed their Youtube website to no longer allow exporting subscriptions using the old method. The only way to export subscriptions now is to use Google Takeout and export them as a .JSON file. Please update your app to import using this format.

gzsombor:
Contributions are always welcomed, unfortunately even if I could work on SkyTube 8 hours a day for the next months, I couldn't implement all the feature requests alone, to address the 138 open issue in this tracker.

profucius:
No worries, I will switch to NewPipe in the meantime, which has resolved this issue in their latest version.

molitona:
> Contributions are always welcomed, unfortunately even if I could work on SkyTube 8 hours a day for the next months, I couldn't implement all the feature requests alone, to address the 138 open issue in this tracker.

We really need that feature It's mandatory to use the app, please.

SebbeM:
I'm looking into this. Parsing JSON should be pretty simple. I'm keeping the XML parsing as well so people with old exports can still use them.

SebbeM:
This is now fixed in PR #878. Please have a look and see what I messed up.

gzsombor:
@SebbeM implementation has been merged

thecountoftuscany:
It seems like Google may have yet again changed their export format: the takeout now gives only a CSV file instead of JSON. The fields in the CSV are: `Channel Id,Channel Url,Channel Title`.

SebbeM:
I can take a look at it but it should probably be treated as a separate issue.

PR #878, opened 2021-01-06
Enable JSON subscription imports

SebbeM:
JSON subscription files can now be imported, as requested in #838. XML files should still work too but I have no file to test with. 

Also, when the channels have been imported, they do not load fully in the list with thumbnail and name. There is just a placeholder thumbnail and an empty space beside it until the channel has been opened. Is this because I missed something or is this old behavior?


gzsombor:
Thanks for your patch! 
* Yes, the problem with missing channel names and icons was already there, because the code, incorrectly just construct a YoutubeChannel object from thin air, and add it to the database, here: https://github.com/ram-on/SkyTube/blob/ff2ab0baa4476c1983de46d5b3009a5a9e0617a5/app/src/main/java/free/rm/skytube/gui/businessobjects/SubscriptionsBackupsManager.java#L342 The proper solution would be to fetch all the channel information from the network - calling https://github.com/ram-on/SkyTube/blob/ff2ab0baa4476c1983de46d5b3009a5a9e0617a5/app/src/main/java/free/rm/skytube/businessobjects/YouTube/newpipe/NewPipeService.java#L239
* I prefer not to touch language files, which I don't speak relatively well - even though removing 'XML-' from the translations everywhere seems too easy and tempting, it could introduce subtle problems. (For example for Hungarian, "Az XML fájl" should become "A fájl" ). If you just change the translation key, weblate will warn all the translators that there are new things to check and translate. So could you please remove the changes in other files?

SebbeM:
Thank you for your input! I will probably get around to reverting the language changes after work today.

SebbeM:
I hope that it was OK for me to fix the Swedish translation. It felt fun as it is my native language.

gzsombor:
Sure, I believe you, that you know Swedish :smile: 
But please, do not modify the other translation files, I see, there are still a lot of `<string name="select_xml_file"> -> <string name="select_sub_file">` changes

SebbeM:
Oh, I'm sorry! I just thought that xml should be removed from the string name and assumed that there would be errors if I didn't fix the references. Should I revert 73287ad and let the references be invalid or should I also change the string name back to say xml again?

gzsombor:
I think the best is to only modify the English and files in your native language. All the other translations could be handled by the translator in Weblate. Having the incorrect/unused translation helps Weblate to show a similar translation, when presenting the new keys.

SebbeM:
I see. I'm not familiar with Weblate so I didn't really know how it works. I'll revert all changes to translations except English and Swedish tomorrow then. Sorry to cause you all this trouble, I just thought I would jump in and do a quick fix, not bother the maintainers! This is my first proper open source contribution so I'm not really used to the procedures, either.

gzsombor:
No problem, I'm thanking you for your work!

SebbeM:
I hope I finally got it right this time! The strings have been changed in the base file and Swedish translation. The name mentioning XML is changed as well as the reference in the Swedish translation file and the code. All other translation files remain untouched.

gzsombor:
Thanks for your first patch! :wink:

PR #879, opened 2021-01-07
A bunch of miscellaneous improvements

TacoTheDank:
The first commit shaves off a couple hundred kilobytes off the APK size of the Extra build flavor.


gzsombor:
Thanks for the patch, looks good!

TacoTheDank:
Thanks!

PR #880, opened 2021-01-08
Dark theme 

sanskar009:


gzsombor:
Could you please rebase it to the latest master? Your branch unfortunately based on older code, and it seems that both branch contains changes in the same files, so automatic rebase is not possible. I tried to understand what went into your branch, but it looks strange - for example this commit: https://github.com/ram-on/SkyTube/pull/880/commits/94cf6dbdf2b735b2019c3fa43e8181ab642522f6  - it looks you cherry-picked a lot of recent change, and this merge: https://github.com/ram-on/SkyTube/pull/880/commits/1f19e4242fe9c176780de9600193a74ab9ee0af5 - but it's hard to follow.

gzsombor:
I can rebase it, and fix the compilation error [here](https://github.com/gzsombor/SkyTube/tree/dark-mode) 
My main issue is that the dark theme does not match the current theme, it has more and different shades of grey. I would rather see that default outlook stays the same. Could you fix it?

drogga:
@sanskar009, @gzsombor - IDK what's going on with this, but I only need the settings (preferences) not to be in White...

drogga:
@sanskar009, @gzsombor - IDK what's going on with this, but I only need the settings (preferences) not to be in White...

gzsombor:
IDK neither, I haven't worked on, after I tried to run, and rebased them. It's not too polished yet.

its0x08 (approved):

drogga:
@hamishmb Forced (from Developer Options or with Shizuku and DarQ) ?, Yes, it is, because that way some elements are dark with dark text, making them unreadable.

hamishmb:
No, just checked and I don't have that option enabled. I'm on Android 11, if that makes any difference?

I haven't noticed anything unreadable - which things display poorly for you?

drogga:
@hamishmb
on S20 FE with A11 and Dark Mode enabled, for some reason the context (3-dots) menu is in white with white text color, making it unreadable.

hamishmb:
For some reason, I don't have that problem on mine. However, I have heard of vendor-dependent Android bugs, so if it affects you, it almost definitely affects others too.

For reference I'm on a Moto G8 Power with Android 11.

drogga:
The whole point here is the settings/preferences menus to be in dark gray (/grey) or true amoled/oled black, along with other UI/layout elements, instead of white, like currently.
I would like to see the line dividers be brought back.

drogga:
The whole point here is the settings/preferences menus to be in dark gray (/grey) or true amoled/oled black, along with other UI/layout elements, instead of white, like currently.
I would like to see the line dividers be brought back.

hamishmb:
Ah yes, I can see the white background when I open preferences, just not in the background of the "..." menu. Agreed that it's a worthwhile addition.

Issue #881, opened 2021-01-08
Bump minSdkVersion to 16/19


TacoTheDank:
Regarding this:
- https://github.com/ram-on/SkyTube/issues/659

And seeing how it seems to be super unstable, I'm wondering if it's time to pull the plug on support <16 (and perhaps even <19).

There's really only so much we can do at this point for such users, and it seems that their user experience is already subpar (with crashes and such). I'd recommend at the very least bumping the minSdkVersion to 16.

gzsombor:
Yeah, it's nearly a year has passed, since I said, that the next release wont support Android 4.0 :smile: And since that, I just postponing the jump - however, there are more compelling reason to do it, as ExoPlayer after 2.9 requires SDK 16, and with that, I could implement subtitle support relatively easily.

drogga:
**Hell No !!!**
Please don't, as I'm truly \ heavily relying on that for \ on couple of devices...

BTW - v2.976 crashes for some reason when i'm trying to play the latest XDA upload (no issues with 2.974 and 2.975):
https://www.youtube.com/watch?v=dtPT5GfIOKI

Also please add a option to disable the update check, or just show it once on every new version and to not show up 'till the next.
Thanks.

To have a support E-Mail address, instead of writing a issue here for absolutely everything will be nice (there's no PMs or anything here, so I hope that you see my point off view) ;)

drogga:
**Hell No !!!**
Please don't, as I'm truly \ heavily relying on that for \ on couple of devices...
(Especially now, since NewPipe Legacy is basically dead - no one cares to update it, also it works worse than this one, plus it's more resource heavy, also I don't have issues on 4.1.2 & 4.3 JB with 2.974 & 2.975)

BTW - v2.976 crashes for some reason when i'm trying to play the latest XDA upload (no issues with 2.974 and 2.975):
https://www.youtube.com/watch?v=dtPT5GfIOKI
Also tested on Android 8 x86 VM\Emulator = the same, so the issue is definitely not in my devices...

Also please add a option to disable the update check, or just show it once on every new version and to not show up 'till the next.
Thanks.

To have a support E-Mail address, instead of writing a issue here for absolutely everything will be nice (there's no PMs or anything here, so I hope that you see my point of view) ;)

@gzsombor @TacoTheDank

thetric:
[...]

PR #883, opened 2021-01-09
Fix parsing of the resolution string (It failed for 1080p60/720p60)

gzsombor:


gzsombor:
Bug mentioned by @drogga in https://github.com/ram-on/SkyTube/issues/881#issuecomment-757122363

PR #928, opened 2021-02-14
README reworked

comradekingu:


gzsombor:
Thanks for the patch! However, it looks like, that the screenshots are not rendered correctly by Github - are you aware of?

comradekingu:
@gzombor Yes, but it should™ work. They were broken before, and then I tried just fixing it like it was, and GitHub wasn't showing the first image correctly.

gzsombor:
Yes, it was broken before, because the images disappeared due to the repository move - I guess.

gzsombor:
Thanks for the patch, after I put the html fragments back, everything looks ok :+1:

PR #929, opened 2021-02-16
Fix Images in the README

gzsombor:


drogga:
@gzsombor There's still 1 broken...

drogga:
@gzsombor There's still 1 broken (called "Browse channel using tablet"), but the next one shows exactly that, so it's a mess there...

Also the GPL 3 one is too large and stretched (pixalated) now.

drogga:
@gzsombor There's still 1 broken (called "Browse channel using tablet"), but the next one shows exactly that, so it's a mess there...

Also the GPL 3 one is too large and stretched (pixalated) now.

gzsombor:
At least, I could fix the GPL image, the other one... This is the original size of the picture, I think

PR #939, opened 2021-03-13
Play / Pause media controls from remote controller

darksworm:
Fixes https://github.com/SkyTubeTeam/SkyTube/issues/936

This Pull Request adds play/pause event handling by implementing `MediaSession` on android 21+ and media buttons `keyDown` handling on android < 21. [Dev docs](https://developer.android.com/guide/topics/media-apps/mediabuttons)

---

### New behavior
When sending play/pause events from remote devices (headphones, remote control, etc) the video playback will be paused / resumed.

---

### Testing
- Virtual devices: Android 9.0, 4.4 - play/pause button in DPAD menu works. 
- Physical devices: Android 8.0 with wireless headset controls work as excpected.
- TV: I couldn't get skytube to work on on my android TV, but I don't see why it wouldn't work.

### DIFF
--- a/app/src/main/java/free/rm/skytube/businessobjects/interfaces/PlaybackStateListener.java
+++ b/app/src/main/java/free/rm/skytube/businessobjects/interfaces/PlaybackStateListener.java
@@ -1,0 +1,7 @@
+package free.rm.skytube.businessobjects.interfaces;
+
+public interface PlaybackStateListener {
+    void started();
+    void paused();
+    void ended();
+}
--- a/app/src/main/java/free/rm/skytube/businessobjects/interfaces/YouTubePlayerFragmentInterface.java
+++ b/app/src/main/java/free/rm/skytube/businessobjects/interfaces/YouTubePlayerFragmentInterface.java
@@ -37,4 +37,6 @@ public interface YouTubePlayerFragmentInterface {
 	 * Start playback of current video.
 	 */
 	void play();
+
+	void setPlaybackStateListener(PlaybackStateListener listener);
 }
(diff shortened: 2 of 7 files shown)


gzsombor:
Thanks for the patch, it looks good! 
Couple of small remarks:
* probably we could handle KeyEvent.KEYCODE_MEDIA_STOP as pause too, what do you think?
* The `void sendPlaybackState(boolean isPlaying, boolean isPaused)` function looks a bit confusing, with the two boolean parameters, as if (isPlaying==true and isPaused==true) could be valid parameters. As they are called only from one place, maybe, if you 'inline' this function, it would make it more simple.

What went wrong with the Android TV? I regularly use SkyTube on one, unless if it's a very old Android, it should work - especially the latest versions needs to work.

darksworm:
> probably we could handle KeyEvent.KEYCODE_MEDIA_STOP as pause too, what do you think?

I don't see why not, I've made the change.

> The void sendPlaybackState(boolean isPlaying, boolean isPaused) function looks a bit confusing, with the two boolean parameters, as if (isPlaying==true and isPaused==true) could be valid parameters. As they are called only from one place, maybe, if you 'inline' this function, it would make it more simple.

Fair enough, I've inlined the method.

> What went wrong with the Android TV? I regularly use SkyTube on one, unless if it's a very old Android, it should work - especially the latest versions needs to work.

I'm not sure, it just doesn't open. I don't even know how to debug it (can't just plug it in to usb and see logs through adb for a TV, right?). Neither the latest public release of skytube, nor my builds worked on it. Not sure if it can be considered old - it is android 8.0.

gzsombor:
Thanks for the patch! I'm going to create a new release with it in the near(*) future.

Hmm, so it's not just the install fails, but when you try to open the app? Strange, indeed. Android 8 should work, it's not that old.
In theory, you can use adb over [wifi](https://developers.google.com/cast/docs/android_tv_receiver/debugging#using_android_debug_bridge), honestly I haven't tried this yet. Unfortunately, to see the logs on the device, at least, you need to run adb once - I mean, you can install [this](https://f-droid.org/en/packages/com.pluscubed.matloglibre/) logviewer, but for accessing the logs, you have to execute an adb command to grant the necessary permissions.

PR #947, opened 2021-03-24
Pause video playback when headphones disconnected

### LINKED ISSUE #435: Playback doesn't pause if headphones disconnect
Please implement pause when headphones (wired or Bluetooth) are disconnected. An option to auto-start the last station on connect would be nice too, if possible, but not as important.

darksworm:
Unplugging wired headphones or disconnecting from bluetooth headphones will pause video playback (fixes #435).

Judging by the docs, this should be compatible starting from API level 3.
These changes will work for both of the player types.

### Behavior

Disconnecting wired or bluetooth headphones:
- while video is playing - video is paused.
- while video paused - nothing happens.
- while video is loading (at 00:00) - nothing happens.

If you have both wired and bluetooth headphones connected and you only disconnect one of them then the video is **not** paused.

### Testing
Tested on a physical Android 9 device with both wired and bluetooth headphones.

Checked disconnecting multiple times in one video, switching videos, letting the phone sleep and returning to the app after a while.

### DIFF
--- a/app/src/main/java/free/rm/skytube/gui/activities/YouTubePlayerActivity.java
+++ b/app/src/main/java/free/rm/skytube/gui/activities/YouTubePlayerActivity.java
@@ -18,8 +18,12 @@
 package free.rm.skytube.gui.activities;
 
 import android.app.Activity;
+import android.content.BroadcastReceiver;
+import android.content.Context;
 import android.content.Intent;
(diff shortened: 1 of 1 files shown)


gzsombor:
Thanks for the patch, it's surprisingly small! :+1:

PR #950, opened 2021-04-01
Use FragmentContainerView for the settings activity

TacoTheDank:
- Add gradle wrapper validation action (see [here](https://github.com/gradle/wrapper-validation-action))
- Update gradle wrapper
- Use FragmentContainerView for the settings activity (see [here](https://developer.android.com/jetpack/androidx/releases/fragment#version_120_3) for reasons for its creation)

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -13,6 +13,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


gzsombor:
Thanks for the patch, it looks great! It's a shame, that for gradle, you need to bring your own binaries - and you need to upgrade them as well - which I found a particularly silly, but that's the way of 'Android' AFAIK :disappointed:

PR #981, opened 2021-07-27
Bump core from 1.3.2 to 1.6.0 and recyclerview to 1.2.1

gzsombor:


TacoTheDank:
I would recommend doing some testing for the Core stuff.

gzsombor:
Yes, testing is always important

PR #982, opened 2021-07-27
Clean up AlertDialogs

TacoTheDank:


gzsombor:
Thanks for the patch!

PR #987, opened 2021-08-04
Use GestureDetectorCompat.

Isira-Seneviratne:
Use [`GestureDetectorCompat`](https://developer.android.com/reference/androidx/core/view/GestureDetectorCompat).


gzsombor:
Thanks for the patch!

PR #988, opened 2021-08-04
Use NotificationChannelCompat.

Isira-Seneviratne:


gzsombor:
Thanks for the patch!

PR #1000, opened 2021-10-22
All app strings reworked

comradekingu:
Hoping to take the app into the beyond #1000 range in various ways :)
I can set up string labels and add screenshots/context if you add
https://hosted.weblate.org/user/kingu/ to https://hosted.weblate.org/projects/skytube//#users
and do some bug triaging on here if need be.
Put in [some issues](https://github.com/SkyTubeTeam/SkyTube/issues/created_by/comradekingu) that I think will help.


gzsombor:
I've found the typo, which broke the build :)
I don't have any special right in weblate, so I can't add you to anything.

PR #1021, opened 2022-02-01
Use newest version of NewPipeExtractor that fixes javascript function…

atomjack:
… name.

For the last day or two I have not been able to view any videos with SkyTube at all, but this update makes them work again. My guess is something at YouTube changed that broke streaming functionality, and luckily NewPipeExtractor was updated to fix this. If others are seeing this problem as well, a new version of SkyTube will need to be released.


gzsombor:
Thanks for the patch!

PR #1048, opened 2022-04-04
Add Sponsorblock

### LINKED ISSUE #892: SponsorBlock support
Hi,
I suggest adding support for [SponsorBlock](https://sponsor.ajay.app/), they have an open API and it was already integrated on Android by the YouTube Vanced project. This would add significantly to SkyTube's appeal.

JamesCullum:
This PR adds opt-in functionality for [SponsorBlock](https://sponsor.ajay.app/).
In the preferences, SponsorBlock can be turned on and the different kind of segments can be enabled or disabled.

> SponsorBlock is an open-source crowdsourced [...] open API for skipping sponsor segments in YouTube videos. Users submit when a sponsor happens from the extension, and the extension automatically skips sponsors it knows about using a [privacy preserving query system](https://github.com/ajayyy/SponsorBlock/wiki/K-Anonymity). It also supports skipping other categories, such as intros, outros and reminders to subscribe, and skipping to the point with highlight.

Closes #892 

[GitHub Sponsors](https://github.com/sponsors/JamesCullum)

**Screenshots**


JamesCullum:
We should consider using [this YouTubeTimeBar library](https://github.com/vkay94/YouTubeTimeBar) in the future - this way we can easily have other features that users are used to from the Youtube app.

I tried to integrate it, but the minimum SDK is 16 (less of an issue) and due to this a lot of things break (more of an issue), which I didn't want to deal with for a first non-invasive commit.

gzsombor:
Nice feature, thanks for working on it! I have small suggestions to improve it a bit.

About SDK 16: well, as I don't own that old devices, I can't really test them, but I know, there are a couple of people who still have them - there was a long discussion about it at: https://github.com/SkyTubeTeam/SkyTube/issues/881
 I think, after implementing some tweaks in upgrade checks for the SkyTube Extra version - to not annoy the Android 4.0 users with incompatible new versions, we could update the SDK version to at least SDK 16. Maybe, we can have a dedicated branch for those old versions, from which F-Droid could still build old releases, parallely to the newer builds.

app/src/main/java/free/rm/skytube/businessobjects/Sponsorblock/SBSegment.java
```diff
@@ -0,0 +1,33 @@
...
+import android.util.Log;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class SBSegment {
+    protected String category;
+    protected double startPos;
+    protected double endPos;
```
gzsombor, line 12: These could be private final fields

app/src/main/java/free/rm/skytube/businessobjects/Sponsorblock/SBTasks.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * A task that retrieves information from the Sponsorblock API about a youtube video
+     *
+     * @param videoId The ID of the youtube video to watch
+     */
+    public static Maybe<SBVideoInfo> retrieveSponsorblockSegments(@NonNull Context context, @NonNull VideoId videoId) {
+        return Maybe.fromCallable(() -> {
+            String[] defaultFilterList = SkyTubeApp.getStringArray(R.array.sponsorblock_filtering_list_default_values);
+            Set<String> filterList = SkyTubeApp.getPreferenceManager().getStringSet("pref_key_sponsorblock_category_list", new HashSet<String>(Arrays.asList(defaultFilterList)));
```
gzsombor, line 40: Could you please move this settings related code to the Settings class? 
Unfortunately not everything located there, but I generally add new things there,

app/src/main/java/free/rm/skytube/businessobjects/Sponsorblock/SBVideoInfo.java
```diff
@@ -0,0 +1,46 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class SBVideoInfo {
+    protected boolean loaded = false;
+    protected double videoDuration;
+    protected List<SBSegment> segments = new ArrayList<SBSegment>();
```
gzsombor, line 13: Apart from the loaded flag, they all can be 'private final'

app/src/main/java/free/rm/skytube/gui/fragments/YouTubePlayerV2Fragment.java
```diff
@@ -151,982 +162,1041 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+
+                    handler.post(new Runnable() {
+                        public void run() {
+                            String rawCategory = payloadSegment.getCategory();
+                            String categoryLabel = rawCategory;
+
+                            try {
+                                Field resourceField = R.string.class.getDeclaredField("sponsorblock_category_" + rawCategory);
+                                int resourceId = resourceField.getInt(resourceField);
+                                categoryLabel = getString(resourceId);
```
gzsombor, line 406: I don't think using reflection is the best solution.
Maybe, you can add a Map<String, Integer> to SBTasks, which maps from category names to resourceIds.
And you can even use that mapping in SBTimeBarView.initDraw to initialize the colors

JamesCullum: I tried that first, but it was a lot of similar code and looked quite ugly. Reflection was the cleanest solution, and it uses a catch with a fallback in case this is incorrect.

app/src/main/java/free/rm/skytube/businessobjects/Sponsorblock/SBVideoInfo.java
JamesCullum: They can be or you want them to be?

JamesCullum: videoDuration also cannot be final (due to being assigned in loop). We could work around that, but thats just more ugly code to reach the same goal.

gzsombor:
Thanks, it looks awesome!

PR #1072, opened 2022-05-17
added dislike counter

andrisas:
added dislike counter using returnyoutubedislike.com API


gzsombor:
Thanks for the patch, nice idea!

Couple of remarks:
* calling a network service from a getter is not the best thing. This could block the UI, or just request the same information over and over again. Could you please move this call to NewPipeService.getDetails? 
* please make it optional, so if someone doesn't want to use that network service, they could opt out with a simple click in the preferences!

gzsombor:
Could you please add a checkbox to a preference page, which could enable/disable this feature?

gzsombor:
Why did you put back the networking call into YoutubeVideo class?

gzsombor:
Thanks for the patch!

drogga:
The dislikes count isn't showing up on the videos from thr feed (subscriptions), this needs a bit of more work...

drogga:
@andrisas, @gzsombor
The dislikes count isn't showing up on the videos from the feed (subscriptions), this needs a bit of more work...

Issue #1046, opened 2022-03-31
NOT ABLE TO IMPORT MY SUBSCRIPTIONS FROM YOUTUBE


lightningzword:
I know that you guys have used NewPipe's API for the backend of the app but this bug is in the NewPipe app too. There too, I am unable to add my subscriptions as it only excepts the ".json" file format. Please look into this ASAP as it is very inconvenient for users. 
Hope to see this fixed in the next update :)

LordJon:
The new way of importing should be possible to implement by using the Google takeout export method. It generates a csv containing all subscriptions (unselect all other YouTube data when exporting).

I discovered SkyTube today in my search for an alternative app in my journey to (partially) DeGoogle. Would love to import my subscription list.

ToXinE:
when I try to use takout importing feature it says quota has been exceeded.

C-O-D:
SkyTube Extra 2.981 still doesn't import abo.csv!
I mean, that's such an important feature, I don't get it...

LordJon:
Yeah I think the devs need some more time as Google completely changed the export format. For now I stopped using the app as manually adding 400 subscriptions isn't realistic.

I'll check this app later again, it had great potential. The add blocking isn't necessary for me though, but I think this is also part of the privacy aspect?

tkalfaoglu:
When I try to import a csv the app just closes

LordJon:
There are multiple reports about this problem which is probably why this one got closed.
Tkalfaoglu, the reason why the app closes is because this problem hasn't been fixed in the public release yet (no clue if there is a fix at all). Google changed the format so importing the csv doesn't work atm.

RedSnt:
Nothing new to add except that I'm really sad this "import youtube subscriptions" feature is broken at the moment, because SkyTube with its SponsorBlock support, looks like the best of adfree youtube players available.

JamesCullum:
Issue is caused by NewPipe and has been fixed upstream: https://github.com/TeamNewPipe/NewPipe/issues/7411

Issue #1041, opened 2022-03-23
Importing subscriptions not working


mclarence554:
I tried to see if exporting the subscriptions in newpipe and then importing to the app to see if this could fix the issue of the subscriptions importing but it also doesnt do anything unfortunately, i thought this could work since newpipe when expoeting the subscriptions it gives a .json file

bobsteva:
Also experiencing this bug, following the guide provided, Google gives me a subscription .csv and not a json. Converting to a json didn't work either, tried multiple formats of .json to no avail. Hopefully this bug gets attention as I can't view my subscriptions because of it unfortunately.

PR #1083, opened 2022-06-19
Refactor Sponsorblock to allow using it offline, "fix" subscription import

JamesCullum:
1. Refactored Sponsorblock, so that when videos are downloaded, segments are displayed and skipped offline as well
2. Removed unused imports (CTRL+SHIFT+O) and auto-indented (CTRL+SHIFT+L) some files I worked with
3. Reworded the "subscription.json" hint to correctly call it "subscription.csv". There were some reports of this not working (#1046 & #1041), but I tested it with the current build and it works. I assume it either was fixed upstream or through a different contribution. Hence we only really need to change the instructions, rebuild and publish to fix this.


gzsombor:
About the imports - there are a couple of file formats which is supported by NewPipeExtractor, json, csv and even zip, and I've fixed a couple of bugs in my [latest commit](https://github.com/SkyTubeTeam/SkyTube/commit/7f977e7ba8eb25a87121b39a655049444a8a601c), but I still received exceptions with the zip format, so I haven't closed the tickets yet.

gzsombor:
Sorry for not replying earlier, I have a couple of concerns:
* generally updating the translations directly could be a pain, as translator could modify those files in weblate, and the only way to interact with that service, is to call "https://hosted.weblate.org/hooks/update/skytube/" - then the weblate service tries to fetch+merge the latest files from this repository. It could fail, then you need to do manually. It's not end of the world, but some extra work is needed. 
* As NewPipeExtractor supports csv/json and zip formats, probably it would be better, to change the label accordingly (sure, the zip file import, maybe still buggy :disappointed: )
* About serializing YoutubeVideo objects into jsons and storing in the database - honestly, I'm not a big fan of that. The problem is, that it's much harder to modify the code, when you need to keep in mind, that YoutubeVideo is also used as a storage format, so we can't add fields which are non-nulls, we can't modify fields type, etc. One other problem, is that with json columns, you can't filter, sort, update the tables. I've already converted the subscriptionVideos table to a more flat structure, so we (or I) can in the future implement searching/filtering/categorization of videos. This is not against your patch, as I see, that you've added a separate column for the SBVideoInfo to mitigate the problems, but this also means that SBVideoInfo is stored twice in that table, once, inside the serialized YoutubeVideo json, and once in a separate column. And as a lesser concern, that it will be also stored in the bookmarks table. I understand, why you've added SBVideoInfo into YoutubeVideo - actually the two are heavily linked, but this makes certain things uglier. For example, when you get the SBVideoInfo from YoutubeVideo, and it's null, you wont know : do we already tried to retrieve the SBVideoInfo from the network? Should we try it again? Maybe, we tried, but the service returned a 404 - or there was a DNS problem, or a connection refused, maybe we won't tried, because that YoutubeVideo comes the database. And there could be also valid questions, if we get back an SBVideoInfo from YoutubeVideo - maybe that info was retrieved couple of days ago, since than, there could be changes, so it could be worth checking again (I'm not sure, if it's a valid scenario that those segments are edited heavily). Maybe instead of putting SBVideoInfo into YoutubeVideo, we can create a wrapper class, which contains: YoutubeVideo, an SBVideoInfo, and some metadata, to distinguish between the cases, when we should try to refetch from the network the latest blocks, and when we shouldn't. What do you think?  
* As the network could potentially block the UI for seconds - or more - could you please add SkyTubeApp.nonUiThread() calls which throws an exception if that happens. So we will notice earlier, if we introduce a potentially blocking behaviour - and we can fix it quickly :wink:

JamesCullum:
Thanks for merging!

I suspected that it would do that, but during my tests this wasnt the case. Maybe you waited for the video to finish completely loading? Anyways, now the issue is fixed either way and the code is more clean this way anyways, thanks for the discussion 👍

Issue #1082, opened 2022-06-18
Ability to set default playback speed


Akv2021:
Problem: I believe a large portion of people (myself included) watch videos at a higher playback speed (1.25x, 1.5x) so it becomes tedious to set it for each new video.

Solution: Ability to set the default playback speed in app preferences (Global setting). Eg. setting it to 1.75x will start playing all the videos from 1.75x speed which the user can then change if needed from the existing playback speed option in video player.

Alternatives: This feature is present in NewPipe, Vanced. Another implementation could be to remember previously used playback speed unless explicitly changed.

Default playback value is expected to persist across app restarts. I shall be happy to provide additional context if needed.

Issue #1068, opened 2022-05-15
ADD global video speed setting 


allanlaal:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

allanlaal:
should not overwrite  #1067 if set for video

ShareASmile:
@gzsombor this issue can be closed as completed in PR #1085

gzsombor:
Thanks @ShareASmile , you are right, this is already released!

PR #1085, opened 2022-06-21
Ability to set default playback speed

andrisas:
Added ability to set default playback speed 
issue #1082 and #1068


gzsombor:
Thanks for the patch! Sorry for the slow response, it looks good. One minor thing, could you please move the settings reading into the Settings class? And please rebase it to the latest main branch.

gzsombor:
Thanks for the patch!

PR #1098, opened 2022-07-20
live videos

andrisas:
added live videos to subs_drawer

### DIFF
--- a/app/src/main/java/free/rm/skytube/gui/fragments/MainFragment.java
+++ b/app/src/main/java/free/rm/skytube/gui/fragments/MainFragment.java
@@ -7,6 +7,7 @@ import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.AutoCompleteTextView;
+import android.widget.Button;
 import android.widget.ImageView;
 
 import androidx.annotation.NonNull;
@@ -25,6 +26,8 @@ import androidx.recyclerview.widget.LinearLayoutManager;
 
 import com.google.android.material.tabs.TabLayout;
 
+import org.schabi.newpipe.extractor.StreamingService;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
@@ -38,8 +41,11 @@ import free.rm.skytube.businessobjects.Logger;
 import free.rm.skytube.databinding.FragmentMainBinding;
 import free.rm.skytube.databinding.SubsDrawerBinding;
 import free.rm.skytube.gui.activities.BaseActivity;
+import free.rm.skytube.gui.activities.MainActivity;
 import free.rm.skytube.gui.businessobjects.adapters.SubsAdapter;
 import free.rm.skytube.gui.businessobjects.fragments.FragmentEx;
+
+import free.rm.skytube.businessobjects.YouTube.newpipe.ContentId;
 
 public class MainFragment extends FragmentEx {
 	// Constants for saving the state of this Fragment's child Fragments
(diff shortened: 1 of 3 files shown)


drogga:
@andrisas This "Live videos" button has no place and shouldn't be there (in the "Subscriptions" side panel), please move it to a tab instead, it's also kind of useless currently, because live videos do stop playing after ~4 seconds...

Request:
There's a reason on why I'm keeping the "Try newer Video formats" option - off (mentioned exactly why many times in diff. issue tickets, the reasons are a couple), but I have to enable it often for videos that do not play otherwise (and live streams - which are being auto-stopped), so can you please add an easy accessible button somewhere in the main UI for that option ?, so I won't have to go to Preferences>Video Player and enable / disable that option constantly, the button can be placed somewhere in the upper right corner, like before or after the "Video Blocker" one (which is useless for/to me and only takes space there), or in the 3-dots drop-down, where the "Enter Video URL" and "Preferences" are placed/located.

Thanks in advance ;)

gzsombor:
Sorry @andrisas if you were waiting for response from me, I was a bit occupied, and as it was marked as draft, I thought, you might be not finished with your changes. 
 Nonetheless, generally I like the idea, and I'm not against having a special link in the subscription slider - some layout adjustment probably needed, and probably a separate tab would be even better - but I totally understand, why haven't you started with that. 
The biggest question, why all the live videos are stopping after a couple of seconds? We must have configure something incorrectly for DASH streams. Or we can say, currently we are just picking up 1-2 URLs and giving it to ExoPlayer and hope the best. Which is clearly not enough. If you look into NewPipe, it's stream selection, downloading, caching mechanism is much-much-much more complex, which we need to copy/imitate somehow. Maybe, we can extract their solution into a library and use it. I mean, the whole network manager, parser and ExoPlayer configurer stuff.  The most optimal would be to convince the NewPipe devs to refactor their code in this way :wink:   
 But I can hardly find any convincing argument, how NewPipe would benefit from that  :crying_cat_face:
 I tried something similar a couple of years ago, but I haven't finished it - mostly, because refactoring a code, which I barely know is much harder, slower and more error prone - and I also feared, that maintaining it - constantly porting bug fixes, enhancements from NewPipe would be an even bigger challenge for.

andrisas:
> The biggest question, why all the live videos are stopping after a couple of seconds? We must have configure something incorrectly for DASH streams.

This is the primary reason why this pr is draft but i am not that familiar with how DASH works

>  Maybe, we can extract their solution into a library and use it. I mean, the whole network manager, parser and ExoPlayer configurer stuff. The most optimal would be to convince the NewPipe devs to refactor their code in this way

I think we could use [this](https://github.com/PierfrancescoSoffritti/Android-YouTube-Player) library instead of using code from newpipe (**i have not tried it**) or we could use [the official youtube android player api](https://developers.google.com/youtube/android/player/) but it needs an api key

gzsombor:
> > The biggest question, why all the live videos are stopping after a couple of seconds? We must have configure something incorrectly for DASH streams.
> 
> This is the primary reason why this pr is draft but i am not that familiar with how DASH works

Neither me, and the ExoPlayer documentation about is not too verbose: https://exoplayer.dev/dash.html


> 
> > Maybe, we can extract their solution into a library and use it. I mean, the whole network manager, parser and ExoPlayer configurer stuff. The most optimal would be to convince the NewPipe devs to refactor their code in this way
> 
> I think we could use [this](https://github.com/PierfrancescoSoffritti/Android-YouTube-Player) library instead of using code from newpipe (**i have not tried it**) or we could use [the official youtube android player api](https://developers.google.com/youtube/android/player/) but it needs an api key

I haven't seen this Android-YouTube-Player, interesting idea to embed a webview with the official player, I have to say! 
My concerns/questions:
* will it play the advertisements too, just like the official website plays? I would rather not see any Youtube Ads, and Sponsorblock would be also nice to keep.
* how customizable can be? And what parts are driven by Youtube? - I mean, based on the github issues, users had problems that the player has changed from one day to another
* Can it also play downloaded videos?

If not, we could keep it only for live videos...

Are you aware, that as of now, SkyTube Extra can embed the official Youtube Player too? I haven't checked that for years, but it's there.

andrisas:
> will it play the advertisements too, just like the official website plays? I would rather not see any Youtube Ads, and Sponsorblock would be also nice to keep.

it will not play ads 

> how customizable can be? And what parts are driven by Youtube? - I mean, based on the github issues, users had problems that the player has changed from one day to another

according to the  [documentation](https://github.com/PierfrancescoSoffritti/Android-YouTube-Player#create-your-own-custom-ui) it can be customized 
> Can it also play downloaded videos?

no

> Are you aware, that as of now, SkyTube Extra can embed the official Youtube Player too? I haven't checked that for years, but it's there.

it works with live videos but it does not support sponsorblock, playback speed changes, like/dislike counter, comments

PR #1123, opened 2022-10-01
Update Android Gradle plugin to 7.3.0.

Isira-Seneviratne:


gzsombor:
Thanks for the patch!

PR #1135, opened 2022-11-12
English short description changed and Russian translation of the short description added. Also fixed icon display

Artem13327:


drogga:
@Artem13327 You can do the same PR for the Legacy repo too...

Also, there are no icons for the 2 on F-Droid, IDK how and who can handle \ fix this...

BTW, the site https://skytube-app.com/ is dead, so the button for "Website" should also be removed from F-Droid and everywhere else too.

drogga:
@Artem13327 You can do the same PR for the Legacy repo too...

Also, there are no icons for the 2 on F-Droid, IDK how and who can and should handle \ fix this and where - on their side or here...

BTW, the site https://skytube-app.com/ is dead, so the button for "Website" should also be removed from F-Droid and everywhere else too.

@IzzySoft, this might concern you too, also nothing comes up when I search for SkyTube in your site, but Google did found it in the repo there, which is a bit strange...

IzzySoft:
> there are no icons

add `fastlane/metadata/android/en-US/images/icon.png` – which is currently missing. I wonder why the `ic_launcher.png` is not picked from the resource directory – but having the icon in Fastlane cannot hurt.

> so the button for "Website" should also be removed from F-Droid

That needs a [MR at fdroiddata](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/12089) (←done). Also removed it for `.extra` in my repo.

> nothing comes up when I search for SkyTube in your site

Uh? That's strange. It's a full-text index behind that which includes the app name column. Let me try:

image

So where did you look?

Meanwhile: That MR passed the pipelines and is merged now. Expect the website link to disappear from the F-Droid listings within a few days (after the next build cycle is through and its results were published).

drogga:
@IzzySoft Thanks, I will let soneone else tag gzsombor, because I think that he doesn't get notified when I tag him for some reason..., so he can merge this and take care of the rest.

I looked here: https://android.izzysoft.de/repo.php , still writes "0 apps found" and I noticed that the icons of the others and their placeholders look strange, some are too big, others too small, maybe 1 app with parge icon breaks the layout for the rest, because all the icons placeholders are too large.

Back to SkyTube: At the same time, as I mentioned - google shows it at: https://android.izzysoft.de/repo/apk/free.rm.skytube.extra, no icon of course, but it's definitely there, don't know why the search doesn't show it up, tho searching for "adb" for example, shows apo results...

IzzySoft:
> I looked here: [android.izzysoft.de/repo.php](https://android.izzysoft.de/repo.php)

I meanwhile consider dropping that "integrated view". Didn't want to use iFrames to integrate the repo there, so I forked out a simple view from the repo itself – which turned out to be hard to keep in sync. Maybe I should add a bold disclaimer linking to the origin – or really remove the tab.

Strange though, as that should use the "old text search" which includes the app name as well.

> maybe 1 app ([android.izzysoft.de/repo/apk/org.adaptit.adaptitmobile](https://android.izzysoft.de/repo/apk/org.adaptit.adaptitmobile)) with large icon (512x512) breaks the layout for the rest

That could indeed be! There was a layout fix for that in the main repo, let me check… Yupp, that wasn't sync'd. As it's only CSS, it was easy to transfer. Can you check again, please?

> All fine here

Yes, as that's what you should use. Guess the easiest thing to do is to replace the search section by a bold hint with a pointer where to go: the original place where the repo resides.

> no icon of course, but it's definitely there

No, it's not. The original repo does not show it either – because the app uses an XML icon only. It *will* however be there once it was found in your fastlane structures.

drogga:
It's better (I think) to just remove that (android.) page \ tab and let google know to not suggest it. Now there's not search field at all there.

I didn't meant that 'there's a icon, but it doesn't show it up', I meant that the SkyTube is indeed there, on the repo, as you wrote too, it seems like you got me wrong there, but NVM, don't worry about it ;)

drogga:
In case (gzsombor) sees this - please take a look at issue 1127 (closed - I'm not mentioning or tagging it properly on purpose), there're some interesting comments by me ;)

IzzySoft:
> and let google know to not suggest it

Go tell them. I've stopped caring about Google (not using it myself). After they told me I shall fix the issue that **they** have pages in their index which they should not according to my `robots.txt`, all I could do was :man_facepalming:

But OK, I've added a `Disallow: /repo` for `User-agent: *` to my `robots.txt`. Let's hope for a purge then.

> I meant that the SkyTube is indeed there

Oh – apologies, misunderstanding then :bow: 

> Yes, I checked, now there's no search field at all there.

That was the easiest "patch" I could think of, without destroying too much usability. The bold text in the header should be clear.

gzsombor:
Thanks for the patch! Sorry for not following the discussion closely - as I'm not familiar with fastlane/f-droid conventions, I can't judge it, but the changes shouldn't be disastrous :wink:

gzsombor:
After some silly typos, I could force push a modified commit...

IzzySoft:
Err, the `icon.png` ended up in the wrong place. Can you please move it into `images/`?

gzsombor:
I hope it's good know

drogga:
^ @gzsombor ^

drogga:
@gzsombor Please add an empty line in the "full_description.txt", as Izzy suggested...

drogga:
@gzsombor Please add an empty line in the "full_description.txt", as Izzy suggested...

PR #1155, opened 2023-02-28
Remove onTerminate() callback override.

Isira-Seneviratne:
Remove the implementation of `onTerminate()`, as it is not called on actual production devices.

Also fixes a crash related to `desugar_jdk_libs` 2.0.

### DIFF
--- a/app/src/main/java/free/rm/skytube/app/SkyTubeApp.java
+++ b/app/src/main/java/free/rm/skytube/app/SkyTubeApp.java
@@ -54,7 +54,6 @@ import com.google.api.client.googleapis.json.GoogleJsonError;
 import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 
 import org.ocpsoft.prettytime.PrettyTime;
-import org.schabi.newpipe.extractor.exceptions.FoundAdException;
 import org.schabi.newpipe.extractor.exceptions.ReCaptchaException;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


gzsombor:
Interesting! Could you tell me, how this onTerminate() override - and the CompositeDisposable usage triggers the original crash? As we use CompositeDisposable in other places, I would assume, that this doesnt change too much, but obviously it seems to fix it

Isira-Seneviratne:
> Interesting! Could you tell me, how this onTerminate() override - and the CompositeDisposable usage triggers the original crash? As we use CompositeDisposable in other places, I would assume, that this doesnt change too much, but obviously it seems to fix it

CompositeDisposable's `add` method calls `Objects.requireNonNull()` to avoid null parameter values being passed.

gzsombor:
I've got that `Objects.requireNonNull()` is called by `CompositeDisposable.add()`, but what I didn't understand, is why this is only problem during the startup - as there are a lot of other `CompositeDisposable.add()` usage through the codebase, but the others don't trigger this failure. So do I understand you correctly, that this happens because SkyTubeApp initialization happens earlier than all the required code is loaded? Strange. 

(As this also happens in the legacy app too, which targets API < 19, probably the linked bug is not relevant - or slightly different)

Isira-Seneviratne:
> I've got that `Objects.requireNonNull()` is called by `CompositeDisposable.add()`, but what I didn't understand, is why this is only problem during the startup - as there are a lot of other `CompositeDisposable.add()` usage through the codebase, but the others don't trigger this failure. So do I understand you correctly, that this happens because SkyTubeApp initialization happens earlier than all the required code is loaded? Strange.

Yeah, it's because multidex is not natively supported on Android 4.4 and older, so the multidex library is needed, and it loads the other dex files during app startup: https://developer.android.com/studio/build/multidex#mdex-gradle

gzsombor:
Thanks for the fix!

PR #1167, opened 2023-04-11
fix screenshots

### LINKED ISSUE #1148: Screenshots in the ReadMe are broken...
No need for a detailed explanation, the screenshots just can't be seen or opened on the main git page.

Can this be fixed?

tomoyanime:
Fix #1148 
You can see how it looks like on my branch https://github.com/tomoyanime/SkyTube/tree/fix-screenshots


gzsombor:
Thank you for the fix!

drogga:
@tomoyanime @gzsombor They still lead to the now dead site when being tapped\clicked on, also in portrait on mobile, the landscape ones appear stretched-out (top to bottom), IDK why or if this last part can be addressed \ fixed somehow...

tomoyanime:
@drogga That's another issue, but I can fix the problems as well. I'll post it as a new issue

drogga:
OK, Thanks. Don't you mean PR (instead of an issue ticket, because they are a lot already, and it might never get fixed by anyone, but be a yet another "stale" one) ?

PR #1174, opened 2023-04-24
New release - v2.986

gzsombor:


drogga:
BTW, pls remove the now dead project's site from the app's about screen\menu, 2 or 3 entries lead to it, it's better if you remove them or just redirect \ replace the links to lead to here (GitHub). Thx.

drogga:
@gzsombor Did you read this, any thoughts, do you care, can you pls at least react so that I know that you saw those comments ? in case you don't want to reply, because it looks like you don't care at all that something implemented very recently simply doesn't work.

PR #1180, opened 2023-04-27
update style of screenshots

tomoyanime:
Ref: https://github.com/SkyTubeTeam/SkyTube/pull/1167
I fixed the aspect ratio issue of screenshots on mobile and updated the links.
You can see how it looks on my branch https://github.com/tomoyanime/SkyTube/tree/optimize-screenshots

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,23 +83,23 @@ You can help us translate this app into your native language by visiting [SkyTub
 ## Screenshots
 ### Phone
 
-<a href="https://skytube-app.com#features">
-<img src="screenshots/trending.jpg" alt="Trending" height="250px"/>
+<a href="screenshots/trending.jpg">
+<img src="screenshots/trending.jpg" alt="Trending" width="140px"/>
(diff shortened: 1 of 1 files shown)


gzsombor:
@tomoyanime thanks for the fix!

drogga:
Hey @gzsombor will you please apply (or transfer) this to the Legacy repo too ?, because IDT there's a need to create the same PRs there too...

Also - Any other way someone (i) can contact you ?, because clearly mentioning \ tagging or pinning you here doesn't work, because it seems like you extremely rarely check your notifications here, or at least don't reply often, so this clearly doesn't work and there are some small but important things that should be tuned a bit and I'm not sure that your Twitter handle is the same (even tho it exists), so pls reply...

Honestly, since there's not PM\DM system here, the best way would probably be if you enable the Discussions tab of your repo fork of ST so I can at least contact you there and I hope that you will be active and actually respond there...

PR #1260, opened 2024-05-04
Cleanup dead invidious instances

drogga:
Feel free to do the same in the Legacy repo


gzsombor:
Thanks for the patch!

drogga:
NP ;)
Please check out my 1st comment in PR 1253 (Closed, I'm not liking it on purpose) and do advice if you would like me to handle some of those or you can do it, because I'm not a coder, so I can so some easy things only and I would like to add more links support, like for Shorts and others from the manifest of NewPipe, but I'm not sure how, I can't really play around much to test and once I create a PR, I can't force push it or do corrections from a browser.
Thank You in advance !

PR #1261, opened 2024-05-06
Fixed typo in cleaner's toast, which was broken

drogga:


drogga:
You migh want to search and replace other similar typos if available, because I think there are, but I'm editing from a browser so I can't.

drogga:
@gzsombor What about this one ?

drogga:
@gzsombor Please merge before the new release because I can't do it myself.

drogga:
@mckaygerhard can you pls. mention/tag gzsombor as wll here to back me up, because I'm afraid that he doesn't get notified when I do it and I don't want this to stay here for even more months and the new ver. if released officially to be without it. Thank You.

gzsombor:
Thanks for the patch, it looks better!

PR #1262, opened 2024-05-06
Added some supported links

drogga:


drogga:
Please check the manifest of the Extra, because I think it has some spacing and lines corrections that should probably be done (not made by me) - it's a bit messy at the beginning.

gzsombor:
Thanks for the patch, sorry for the long delay!

Yes, there is one or two missing new lines, thanks for noticing it, I'm going to fix it.

PR #1283, opened 2024-07-15
Exported the preferences activity in the main AndroidManifest.xml

drogga:


drogga:
@gzsombor Let's hope that you will merge this soon, instead of taking a few months...

gzsombor:
Sorry for the usual delays, again, my bad!

What do you want to achieve with this change? Based on my knowledge - https://developer.android.com/guide/topics/manifest/activity-element#exported - that flag is only needed to set to true, if you want to trigger this activity from other programs? Is there some specific scenario, where this setting is beneficial for the preference activity?

drogga:
Yes!

drogga:
@gzsombor
https://github.com/SkyTubeTeam/SkyTube/blob/36655ba4210798373a34c5875e7b79d1d2d25d86/app/src/main/AndroidManifest.xml#L58 is this really necessary and does it actually do anything in this repo, especially given that the minSDKversion if this one is 4.4 KitKat (hugher) ? (there are also other places mentioning JB in code)

drogga:
@gzsombor > Yes.
___
https://github.com/SkyTubeTeam/SkyTube/blob/36655ba4210798373a34c5875e7b79d1d2d25d86/app/src/main/AndroidManifest.xml#L58 is this really necessary and does it actually do anything in this repo, especially given that the minSDKversion of this one is 19 = 4.4 KitKat (hugher than JB) ? (there are also other places mentioning JB in code)

gzsombor:
> @gzsombor > Yes.
> 

Sorry, I still don't get, what is your goal to have this flag set to true?

> https://github.com/SkyTubeTeam/SkyTube/blob/36655ba4210798373a34c5875e7b79d1d2d25d86/app/src/main/AndroidManifest.xml#L58
> is this really necessary and does it actually do anything in this repo, especially given that the minSDKversion of this one is 19 = 4.4 KitKat (hugher than JB) ? (there are also other places mentioning JB in code)

Based on a quick google search for the [documentation](https://developer.android.com/studio/write/tool-attributes) you are right, these tools:targetApi annotations are not doing too much

drogga:
It won't hurt if the preferences activity is exported and launchable without root by other apps, actually it doesn't makes any sense to not be exported, why would it be, what does it achieve - nothing, but being exported I can use not only a shortcut, but an automation software (like Tasker) to trigger some things, without root, which I don't have on my daily driver. It would be too much complicated and long to explain what I want to do, just because I've changed an activity state with a param/attribute/flag.

PR #1345, opened 2025-03-02
docs: structural and grammatical improvements

RightWrite:
Regarding the use of commas at the end of list items: There is flexibility in vertical list formatting. Some style guides recommend using semicolons after bulleted list items that are long & contain commas (as this list does). However, these are suggestions & not requirements. I removed commas at the end of list items because they are unnecessary & deplete aesthetics. 

F-Droid supports the unordered list tag `<ul></ul>`, so I placed the list items within those tags in this file. If, for whatever reason, you want to keep the asterisks, that's fine — just let me know & I'll make that change.

Some sentences were restructured for readability & clarification. 

Regarding the improvement of line 2 "_block unwanted videos_." To me this seems a bit obtrusive because the statement is rather large compared to the rest of the list items. There are a couple of options: One is to use a nested vertical list; another is to rewrite the phrase. In this case I converted it into a sentence & placed it toward the end of the description. I think this also enhances the look of the description. If you want to go the nested list route (or have other suggestions), just let me know. 

Redundant words were removed. Some words were unnecessary because the context of the phrase made it obvious what you are referring to. For instance, on line 3 "_and video description_" the word "_video_" was removed; your audience knows that you're talking about videos in this phrase because it is mentioned at the start. 

The feature list was rearranged to enhance logical flow.

### DIFF
--- a/fastlane/metadata/android/en-US/full_description.txt
+++ b/fastlane/metadata/android/en-US/full_description.txt
@@ -1,19 +1,22 @@
 SkyTube is a YouTube player that allows you to:
-* block unwanted videos (including channel blacklisting/whitelisting, language filtering, low view video blocking, high dislikes video blocking... etc.),
-* video swipe controls (including volume, brightness, comments and video description controls),
-* bookmark videos,
-* import subscriptions from YouTube,
-* play channels' playlists,
-* download videos,
-* view and download video thumbnails
-* explore Trending and Most Popular videos,
-* browse YouTube channels,
-* play YouTube videos,
-* view video comments,
-* search videos, music and channels,
-* channel subscription & non-intrusive notifications,
-* subscriptions feed
+<ul>
+<li>play YouTube videos</li>
+<li>download videos</li>
+<li>browse YouTube channels</li>
+<li>search videos, music and channels</li>
+<li>watch channels' playlists</li>
+<li>view video comments</li>
+<li>video swipe controls (volume, brightness, comments and description)</li>
+<li>bookmark videos</li>
+<li>import subscriptions from YouTube</li>
+<li>channel subscription with non-intrusive notifications</li>
+<li>subscriptions feed</li>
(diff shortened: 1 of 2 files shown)


gzsombor:
Thanks for your contribution, I didn't know that you can write html in those txt files, great!
 It looks great, the only minor thing is that in the 
 "An AD-Free and Open Source YouTube Client~" 
The AD looks strange, isn't it written as 'An Ad-free ...'?

RightWrite:
You are welcome!

Yes sir, the phrase ‘AD-Free’ does seem a bit odd. 

Nice catch! 

For some reason I forgot to include a message about the _short_description_ in the PR. Here it is: `Regarding the short_description.txt file: The word ‘ad-free’ was inserted & the text was converted to title case (all major words capitalized) that optimally represents a fragmented phrase.`

Because it’s a subtitle I converted the phrase to _title case_. And to emphasize that it was a fragmented sentence (with title caps) I placed a tilde on each side.

However, I will place it all in lower case or _sentence case_ (only the first word capitalized) if you like. Along those lines, I can remove the tildes on the sides & place terminal punctuation (a period) at the end, to keep it tidy. But, again, I won’t add that if you don’t want. 

I can take care of it this evening. 

Let me know how you want to proceed.

thanks

RightWrite:
To clarify, the initial commit didn't make sense because the reason for the change (which I later included) was missing. 

In this version it's been converted into a sentence. 

Note: The compound word "open-source" is hyphenated because it describes the noun "client." However, I will remove the hyphen it you want. 

Let me know what you think.

gzsombor:
Thank you for your help and explanation! :smile:

RightWrite:
You are welcome! 

Glad to help out. :smile: 

I found a mistake in the full_description file, so I will rewrite it & send another PR. 

That will get done today.

thanks,

PR #1370, opened 2025-08-27
Security pin

lexarvn:
Add security pin that is needed for changing preferences and blocking/unblocking channels

Possible fix for https://github.com/SkyTubeTeam/SkyTube/issues/774


lexarvn:
I pushed some changes fixing issues copilot pointed out

gzsombor:
Thanks for your contribution!

Issue #1391, opened 2026-01-10
Original YouTube video in  Italian, in your player in English


conant69:
**Describe the bug**

Original YouTube video in Italian, in your player the language become in English, it's happen very frequently.

Thank you

Issue #1382, opened 2025-10-10
Vidéos françaises doublées en anglais


fredericclarysse-cpu:
sont lues avec le doublage automatique en anglais..

maaaat:
There's a setting for that: `Video and audio` → `Prefer original audio`.

Honestly this option should be turned on by default, this slop is getting obnoxious. Another related thing is that video titles are sometime auto-translated, and I haven't seen an option in SkyTube to always display the original text.

Edit: my apologies I am confusing SkyTube and NewPipe, since I use both apps. The option I mention here is for NewPipe, not SkyTube.

jgmy:
I have some Issue. Will try that "original audio" option.
(Oh, that option is not in my version).
Just in case someone want videos dubbed, but not for English, could you add a setting to send the locale string along with the http request?

Issue #1371, opened 2025-08-30
Auto dubb issue


kaladharan7930-cell:
App doesn't support audio switching,by default it only support audio youtube recommend which may not original audio.
         I found nothing to change option

drogga:
> There's now a setting for that: `Video and audio` → `Prefer original audio`.

**No, there isn't**, you probably aren't even using this (SkyTube) app, but NewPipe, there there's such option, not here unfortunately, please remove your misleading comments (incl. in 1382, or edit it). Thank You.

Similar tickets: #1323 #1369 #1382 #1391

Issue #1369, opened 2025-08-25
audio tracks error


jobbex:
**Describe the bug**
When I open a YouTube video with multiple audio tracks, even though I've set the SkyTube app language to Italian and my tablet always has the same language, the videos are played in English. I should point out that if I open the same video from a browser, I hear it directly in Italian. For example
https://www.youtube.com/shorts/nZnKymTaWNo

**Expected behavior**
I would like to hear the audio in Italian

**Screenshots**
No

**Setup:**
 - Android version: 7.0
 - Is Your device rooted? No
 - Do You use Tor or any VPN service with your device? No
 - App source: FDroid
 - App Version: 2.995
 - App Flavour 

Thanks

tinman2501:
Very similar issue on Android 15.

I even managed to confuse the app to get videos translated into different languages.

PR #1411, opened 2026-04-21
AI dubbing issue solved

ialnavy:
This pull request solves: #1371 #1323 #1369 #1382 #1391

## Description

When reproducing a video, it may sometimes use the AI dubbing audio track to a random language.

## Analysis

I have been reproducing this issue by reproducing videos by their URL. I noticed that it is sensitive to the "Try newer Video formats" option. Disabling this option solves the issue with the audio, but it reproduces a low quality video, due to the system using the legacy video sources.

The bug is in the "pickAudio" method of class "StreamSelectionPolicy". When searching for the best audio stream, by invoking "isBetter", it does not discriminate AI dubbing tracks.

## Solution

I added an auxiliary method "isOriginalAudio", which only accepts original streams ([AudioTrackType](https://pub.dev/documentation/android_new_pipe_extractor/latest/android_new_pipe_extractor/AudioTrackType.html).ORIGINAL) or legacy streams (null).

The "pickAudio" method will search, not only for the best audio stream, but also for an original one.

Having disabled the "Try newer Video formats" option worked because it was forcing the system to take the legacy audio stream.

See commits for further info.

## Considerations for the future

It should be considered, developing a proper option "Use AI dubbing", as well as a combo box for setting an explicit preferred language for AI dubbing.

### DIFF
--- a/app/src/main/java/free/rm/skytube/app/StreamSelectionPolicy.java
+++ b/app/src/main/java/free/rm/skytube/app/StreamSelectionPolicy.java
@@ -22,6 +22,7 @@ import android.net.Uri;
 
 import org.schabi.newpipe.extractor.MediaFormat;
 import org.schabi.newpipe.extractor.stream.AudioStream;
+import org.schabi.newpipe.extractor.stream.AudioTrackType;
 import org.schabi.newpipe.extractor.stream.StreamInfo;
 import org.schabi.newpipe.extractor.stream.VideoStream;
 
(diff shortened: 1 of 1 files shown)


gzsombor:
Thanks for your contribution, that looks like a good idea!

DanielGlaas:
> This pull request solves: #1371 #1323 #1369 #1382 #1391

Only one oft these five issues is closed. Was this forgotten?

